PR #36, opened 2017-07-24
Support Icons/Glyphs

pkanuganti:
Support Icons/Glyphs
[Story](https://www.pivotaltracker.com/story/show/148395153)


packages/react-ui-core/src/Icon/Icon.js
```diff
@@ -0,0 +1,41 @@
+import React, { Component } from 'react'
+import RPT from 'prop-types'
+
+export default class Icon extends Component {
```
tadjohnston: If this is only supporting SVG file, I really don't think it should be called `Icon` as the component name. @lfender6445 thoughts?

packages/react-ui-core/src/Icon/Icon.js
```diff
@@ -0,0 +1,41 @@
+import React, { Component } from 'react'
+import RPT from 'prop-types'
```
tadjohnston: Please import this as `PropTypes` to keep consistency with the other components

packages/react-ui-core/src/Icon/Icon.js
```diff
@@ -0,0 +1,41 @@
...
+      RPT.object,
+    ]),
+  };
+
+  static defaultProps = {
+    color: '#000',
+    size: 30,
+  }
+
+  render() {
```
tadjohnston: Please put the `render` method last.

packages/react-ui-core/src/Icon/Icon.js
```diff
@@ -0,0 +1,41 @@
...
+    const { wrapperStyle } = this.props
+    if (wrapperStyle) { return <div style={wrapperStyle}>{this.getIcon()}</div> }
+
+    return this.getIcon()
+  }
+  getIcon() {
+    const { color, height, onClick, size, style, width, pathD, viewBox } = this.props
+    return (<svg fill={color} height={height || size} width={width || size} onClick={onClick}
```
tadjohnston: The props should be on separate lines. Please fix formatting, you can refer to other components.  Also, I think this is way too many props for this component.

packages/react-ui-core/src/Icon/Icon.js
```diff
@@ -0,0 +1,41 @@
+import React, { Component } from 'react'
+import RPT from 'prop-types'
+
+export default class Icon extends Component {
+
+  static propTypes = {
+    color: RPT.string,
+    height: RPT.number,
+    pathD: RPT.string.isRequired,
```
tadjohnston: What is `pathD`?

pkanuganti: @tadjohnston pathD is nothing but svg-path tag data.

packages/react-ui-core/src/Icon/Icon.js
pkanuganti: Modified and checked-in.

packages/react-ui-core/src/Icon/Icon.js
pkanuganti: modified and committed

packages/react-ui-core/src/Icon/Icon.js
pkanuganti: modified and committed

packages/react-ui-core/src/Icon/Icon.js
lfender6445: prvaeen, please investigate ag.js implementation, is it specific to svgs

packages/react-ui-core/src/Icon/Icon.js
```diff
@@ -0,0 +1,51 @@
...
+            <path
+                d={pathD}
+            />
+        </svg>
+    )
+  }
+  render() {
+    const { wrapperStyle } = this.props
+    if (wrapperStyle) {
+      return <div style={wrapperStyle}>{this.getIcon()}</div>
```
lfender6445: rather than support an inline style and because this wraps a primitive html element, why don't we pass along the props llike this https://github.com/rentpath/react-ui/blob/master/packages/react-ui-core/src/Form/Input.js#L60 for full control

stories/utils/pathd.js
```diff
@@ -0,0 +1,3 @@
+export const SETTINGS = 'M542.2-29.2l-2.3-.8-9.1-21.1 1-2.2c27.7-60.7 25.7-62.5 21.1-66.9L516-155c-1.5-1.5-4.1-2.4-6.4-2.4-2.1 0-8.4 0-62.7 23.8l-2.2 1-22.4-8.7-.8-2.2C395.8-205 393.2-205 386.8-205h-51.9c-6.3 0-9.2 0-32.9 61.5l-.8 2.3-22.3 8.9-2.2-.8c-36.8-15-58.2-22.7-63.8-22.7-2.3 0-5 1-6.4 2.4l-36.9 34.9c-4.6 4.5-6.6 6.3 22.4 65.8l1.1 2.2-9.2 21.1-2.2.8c-64.8 24.3-64.8 26.7-64.8 33v49.4c0 6.3 0 9 64.9 31.5l2.3.8 9.2 21-1 2.1c-27.5 60.7-25.8 62.4-21.1 66.9l36.8 34.8c1.6 1.5 4.2 2.4 6.4 2.4 2.1 0 8.4 0 62.7-23.8l2.2-1 22.4 8.9.8 2.2c25.6 61.4 28.1 61.4 34.6 61.4H389c6.6 0 9.2 0 32.9-61.5l1-2.3 22.3-8.7 2.1.8c36.8 15 58.2 22.7 63.8 22.7 2.3 0 5-1 6.4-2.4l36.9-34.9c4.6-4.5 6.6-6.3-22.6-65.6l-1.1-2.2 9.1-21 2.2-.8c65-24.5 65-26.9 65-33.2V2.5c.1-6.5.1-9.2-64.8-31.7zM362 108.1c-46.5 0-84.2-35.9-84.2-80.1s37.9-80.1 84.2-80.1 84.2 35.9 84.2 80.1-37.8 80.1-84.2 80.1z'
```
lfender6445: what if we stored viewbox info here as well as an object? like kv pair 

```
SETTINGS_ICON= {
  viewBox: '', 
  pathD: '', 
}
```

`<Icon {...icon.settings} />`

packages/react-ui-core/src/Icon/Icon.js
pkanuganti: Modified and pushed changes.

stories/utils/pathd.js
pkanuganti: I feel, its better to have independent setting, because for the scenario like this, user 1 & user 2 want have same icon with different viewbox size. if we map together, we need to maintain two settings for same icon. Let us know your thoughts, we can change accordingly.

packages/react-ui-core/src/Icon/Icon.js
pkanuganti: I have gone through the ag.js. It is specific to svg

tadjohnston: That may be true, but that may not be the case on rent-js or any of our other sites.

packages/react-ui-core/src/Svg/Svg.js
```diff
@@ -0,0 +1,29 @@
...
+    className: PropTypes.string,
+  }
+
+  render() {
+    const { pathD } = this.props
+    return (
+      <svg {...this.props} >
+        <path d={pathD} />
+      </svg>
+    )
```
tcwilson: @lfender6445 has there been any discussion on this way of handing svg icons?

To me, this doesn't seem to be of any benefit. You may as well just write `<svg ...>` directly. It also limits us to just using `path`, and a single one at that. Take [this svg icon](https://raw.githubusercontent.com/rentpath/ag.js/master/src/apps/shared/Icon/icons/arrow_down.svg?token=AAOby1kduMYgJDWSVYOf6VsvxR5MfcpLks5ZghDHwA%3D%3D) for example. Or [this one](https://raw.githubusercontent.com/rentpath/ag.js/master/src/apps/shared/Icon/icons/crosshair.svg?token=AAOby9g1HEo6y1tI-jwtKDAoFJu4Ceobks5ZghFzwA%3D%3D).

Maintaining a bunch of path strings in JS seems like a maintenance headache and a step back from what AG.js does, with SVG sprites and `xlink:href`

lfender6445: yeah we talked about it this yesterday and will be closing it out

PR #37, opened 2017-07-25
Changes as per the requirements for Inc_Dec_Filter

saurabhuideveloper:
[story](https://www.pivotaltracker.com/story/show/148394571)
Hey guys I have created this new branch for Increment Decrement Filter so want your review on this.


packages/react-ui-core/src/IncrementDecrementFilter/IncrementDecrementFilter.js
```diff
@@ -0,0 +1,53 @@
...
+      })
+    }
+  }
+  render() {
+    const { count } = this.state
+    const { theme, label, leftUnit, rightUnit, id } = this.props
+    return (<div id={id} className={theme.container}>
+      <div className={theme.label}>{label}</div>
+      <div className={theme.componentContainer}>
+          <span onClick={this.increment} className={`leftUnit ${theme.contentInnards} ${theme.circle}`}>&#x002B;</span>
```
HansHauge: We should probably break into new lines 

```JavaScript
<span 
  onClick={this.increment} 
  className={`leftUnit ${theme.contentInnards} 
  ${theme.circle}`}>
    &#x002B;
</span>
```

HansHauge:
👍

packages/react-ui-core/src/IncrementDecrementFilter/IncrementDecrementFilter.js
```diff
@@ -0,0 +1,53 @@
+import React, { Component, PropTypes } from 'react'
```
tadjohnston: Please import PropTypes directly from the lib itself. It will eventually be deprecated from the react package.

```
import PropTypes from `prop-types`
```

packages/react-ui-core/src/IncrementDecrementFilter/IncrementDecrementFilter.js
```diff
@@ -0,0 +1,53 @@
...
+  }
+  render() {
+    const { count } = this.state
+    const { theme, label, leftUnit, rightUnit, id } = this.props
+    return (<div id={id} className={theme.container}>
+      <div className={theme.label}>{label}</div>
+      <div className={theme.componentContainer}>
+          <span onClick={this.increment} className={`leftUnit ${theme.contentInnards} ${theme.circle}`}>&#x002B;</span>
+          <span className={`${theme.contentInnards} ${theme.count}`}>{`${leftUnit} ${count} ${rightUnit}`}</span>
+          <span onClick={this.decrement} className={`rightUnit ${theme.contentInnards} ${theme.circle}`}>&#x002D;</span>
```
tadjohnston: The `+` and `-` children on the spans should probably be customizable.  Having them as defaults is fine, but the should allow for an arbitrary component / node.

packages/react-ui-core/src/IncrementDecrementFilter/IncrementDecrementFilter.js
```diff
@@ -0,0 +1,53 @@
+import React, { Component, PropTypes } from 'react'
+
+export default class IncrementDecrementFilter extends Component {
```
tadjohnston: I'd like to rename this component to `Counter`.  @lfender6445 what do you think? The name currently is a little too long IMO and it's really _not_ a filter either. Yes, technically on rent-js it is, but on react-ui it should be agnostic to what it's being used for.

lfender6445: yes i like Counter

packages/react-ui-core/src/IncrementDecrementFilter/IncrementDecrementFilter.js
```diff
@@ -0,0 +1,69 @@
...
+    const { count } = this.state
+    const { theme, label, leftUnit, rightUnit, id } = this.props
+    return (
+      <div id={id} className={theme.container}>
+        <div className={theme.label}>
+          {label}
+        </div>
+        <div className={theme.componentContainer}>
+          <span
+            onClick={this.increment}
```
lfender6445: we should allow onClick to pass through the component for example 

``` 
  const count =  count
  if (this.props.onClick) { 
    this.props.onClick(count)
  }
  this.setState({
     count: decrementValue,
  })
```

packages/react-ui-core/src/IncrementDecrementFilter/IncrementDecrementFilter.js
```diff
@@ -0,0 +1,69 @@
...
+            className={`leftUnit ${theme.contentInnards} ${theme.circle}`}
```
lfender6445: i am not sure i like these classnames, we should not assume a style or flavor of button here 

for example `theme.circle` implies that this component will always be a circle

we want things generic as possible, i will explain more in the morning. 

also, please use the `classnames` library for concatting classes so if a theme  does not have it it is does not add a class of undefined.

packages/react-ui-core/src/IncrementDecrementFilter/IncrementDecrementFilter.js
```diff
@@ -0,0 +1,69 @@
...
+            &#x002B;
+          </span>
+          <span
+            className={`${theme.contentInnards} ${theme.count}`}
+          >{`${leftUnit} ${count} ${rightUnit}`}</span>
+          <span
+            onClick={this.decrement}
+            className={`rightUnit ${theme.contentInnards} ${theme.circle}`}
+          >
+            &#x002D;
```
lfender6445: regarding these html entities we should not be making assumptions about `+` or `-` signs although they are good defaults I would prefer it to be more generic and instead allow the user to inject their own button

packages/react-ui-core/src/Counter/Counter.js
```diff
@@ -10,18 +12,10 @@ export default class IncrementDecrementFilter extends Component {
...
+    this.props.leftOperatorClick(this)
   }
   decrement() {
-    if (this.state.count > 1) {
-      const decrementValue = this.state.count - 1
-      this.setState({
-        count: decrementValue,
-      })
-    }
+    this.props.rightOperatorClick(this)
```
lfender6445: why not use a single binding to handle the click?

packages/react-ui-core/src/Counter/Counter.js
```diff
@@ -34,36 +28,43 @@ export default class IncrementDecrementFilter extends Component {
...
 
-IncrementDecrementFilter.defaultProps = {
+Counter.defaultProps = {
   theme: {},
   count: 1,
   leftUnit: '',
   rightUnit: '',
+  leftOperatorClick: '',
+  rightOperatorClick: '',
+  leftOperator: <span>?</span>,
```
lfender6445: why use a question mark instead of something like a plus or minus

packages/react-ui-core/src/Counter/Counter.js
```diff
@@ -34,36 +28,43 @@ export default class IncrementDecrementFilter extends Component {
...
 }
 
-IncrementDecrementFilter.defaultProps = {
+Counter.defaultProps = {
   theme: {},
   count: 1,
   leftUnit: '',
   rightUnit: '',
+  leftOperatorClick: '',
+  rightOperatorClick: '',
```
lfender6445: this is a blank string, but its of type function so it doesn't make sense to use a string here

packages/react-ui-core/src/Counter/__tests__/Counter-test.js
```diff
@@ -0,0 +1,68 @@
+import React from 'react'
+import { shallow } from 'enzyme'
+import Counter from '../Counter'
+
+const theme = {}
+const leftOperatorClick = function (component) {
+  const incrementValue = component.state.count + 1
+  component.setState({
```
tcwilson: Calling `setState` should only be done in the component, itself.

If you want to decouple the actual count from the `Counter` component then you should pass in a `count` prop and have `Counter` render using that, and keep the `count` state outside of the component. However, I can't think of a reason you would really need to specialize your increment/decrement to do anything other than go down/up by one. Maybe there is a use case there, but this may be a case of [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it).

What I would probably do is do the `setState` in the `increment()` or `decrement()` of `Counter` component. This way, the user of `Counter` doesn't have to implement state handling and the component is useful right away. On top of that, you can easily add an `onIncrement` or `onDecrement` prop callback to override the default behavior later on.

packages/react-ui-core/src/Counter/Counter.js
```diff
@@ -0,0 +1,69 @@
...
+    const { count } = this.state
+    const { theme, label, leftUnit, rightUnit, id } = this.props
+    return (
+      <div id={id} className={theme.container}>
+        <div className={theme.label}>
+          {label}
+        </div>
+        <div className={theme.componentContainer}>
+          <span
+            onClick={() => this.onClick('increment')}
```
lfender6445: For increment and decrement we only care about the numeric value, not whether it went up or down. Lets modify the click event to just pass the number rather than a string: 

`onClick={() => this.onClick(this.state.count + 1)})`

rename onClick -> handleClick and make sure the props.onClick exists before calling it: 

```
handleClick(count) {
  if (this.props.onClick) {
    this.props.onClick(count)
  }
  this.setState(count)
}
```

packages/react-ui-core/src/Counter/Counter.js
```diff
@@ -0,0 +1,69 @@
...
+  }
+}
+
+Counter.defaultProps = {
+  theme: {},
+  count: 1,
+  leftUnit: '',
+  rightUnit: '',
+  handleClick: () => {
+    alert('please add rightOperatorClick handler function')
```
tadjohnston: This should _not_ be here.  If this is required, it should be required in `Counter.propTypes` or you can use a noop here instead.

packages/react-ui-core/src/Counter/Counter.js
```diff
@@ -0,0 +1,69 @@
...
+  },
+  leftOperator: <span>-</span>,
+  righttOperator: <span>+</span>,
+}
+
+Counter.propTypes = {
+  leftOperator: PropTypes.element,
+  righttOperator: PropTypes.element,
```
tadjohnston: I don't think we should label things `leftOperator, `righttOperator` (which is spelled wrong here as well), `leftUnit` or `rightUnit`.  If the increment / decrement buttons are on the top and bottom these names would be irrelevant.  I think naming them `increment` or `decrement` or something would be better.

packages/react-ui-core/src/Counter/Counter.js
```diff
@@ -0,0 +1,69 @@
...
+          </span>
+        </div>
+      </div>
+    )
+  }
+}
+
+Counter.defaultProps = {
+  theme: {},
+  count: 1,
```
aribouius: Why would this default to `1` instead of `0`?

packages/react-ui-core/src/Counter/Counter.js
```diff
@@ -0,0 +1,69 @@
...
+    this.props.handleClick(count)
+    this.setState({
+      count,
+    })
+  }
+  render() {
+    const { count } = this.state
+    const { theme, label, leftUnit, rightUnit, id } = this.props
+    return (
+      <div id={id} className={theme.container}>
```
aribouius: Why are we explicitly supporting an `id` attribute?

packages/react-ui-core/src/Counter/Counter.js
```diff
@@ -0,0 +1,69 @@
...
+            className={classNames(theme.contentInnards, theme.count)}
+          >{`${leftUnit} ${count} ${rightUnit}`}</span>
+          <span
+            onClick={() => this.handleClick(this.state.count - 1)}
+            className={classNames('rightUnit', theme.contentInnards, theme.shape)}
+          >
+            {this.props.rightOperator}
+          </span>
+        </div>
+      </div>
```
aribouius: The theme classnames here don't follow our CEM (component-element-modifier) convention.  I would expect the classnames to look more like:
```
.Counter {} // component
.Counter-active // component-modifier
.Counter_label {} // component-element
.Counter_label-active {} // component-element-modifier
```

lfender6445:
please remove package-lock.json and use yarn install for installing dependencies not npm

packages/react-ui-core/src/Counter/Counter.js
```diff
@@ -0,0 +1,67 @@
...
+    const { theme, label, decrementUnit, incrementUnit } = this.props
+    return (
+      <div className={theme.Counter}>
+        <div className={theme['Counter-label']}>
+          {label}
+        </div>
+        <div className={theme['Counter-label-active']}>
+          <span
+            onClick={() => this.handleClick(this.state.count + 1)}
+            className={classNames('incrementUnit', theme['Counter-content'], theme['Counter-shape'])}
```
lfender6445: because we use locally scoped css modules we should not hard code classes in any of the components. lets find a better way to target this item here and in your tests

packages/react-ui-core/src/Counter/Counter.js
```diff
@@ -0,0 +1,67 @@
...
+    )
+  }
+}
+
+Counter.defaultProps = {
+  theme: {},
+  count: 0,
+  decrementUnit: '',
+  incrementUnit: '',
+  handleClick: function noop() {},
```
tadjohnston: You can just do this instead which would be better:

```
handleClick: () => {}
```

packages/react-ui-core/src/Counter/Counter.js
```diff
@@ -0,0 +1,67 @@
...
+            onClick={() => this.handleClick(this.state.count + 1)}
+            className={classNames('incrementUnit', theme['Counter-content'], theme['Counter-shape'])}
+          >
+            {this.props.decrementOperator}
+          </span>
+          <span
+            className={classNames(theme['Counter-content'], theme['Counter-count-content'], theme['Counter-Shape'])}
+          >{`${decrementUnit} ${count} ${incrementUnit}`}</span>
+          <span
+            onClick={() => this.handleClick(this.state.count - 1)}
```
tadjohnston: What if I want to decrement / increment by something like say 2? How do I do that?

packages/react-ui-core/src/Counter/Counter.js
```diff
@@ -0,0 +1,71 @@
...
+        <div className={theme['Counter-label-active']}>
+          <span
+            role='incrementUnit'
+            onClick={() => this.handleClick(this.state.count + changeValue)}
+            className={classNames(theme['Counter-content'], theme['Counter-shape'])}
+          >
+            {this.props.decrementOperator}
+          </span>
+          <span
+            className={classNames(theme['Counter-content'], theme['Counter-count-content'], theme['Counter-Shape'])}
```
tadjohnston: Nitpick: Can we put these classes on separate lines so it's easier to read? (same across the file)

lfender6445: done. merging

PR #38, opened 2017-07-28
Export Shared AG.js Map To React-UI

maityarkodeb:
[Story](https://www.pivotaltracker.com/story/show/148992455)


packages/react-ui-core/src/Map/Map.js
```diff
@@ -0,0 +1,58 @@
...
+      color,
+      theme,
+      className,
+      ...props
+    } = this.props
+
+    return (
+      <div
+        {...props}
+        style={{
```
lfender6445: remember, react components should be agnostic of styles

packages/react-ui-core/src/Map/Map.js
```diff
@@ -0,0 +1,58 @@
...
+
+  static defaultProps = {
+    theme: {},
+  }
+
+  componentDidMount() {
+    mapboxgl.accessToken = this.props.token
+    const map = new mapboxgl.Map({
+      container: 'map',
+      style: 'mapbox://styles/mapbox/streets-v9',
```
lfender6445: we should not be hardcoding the map styles, these components should be flexible enough to be customized from the outside in.

packages/react-ui-core/src/Map/Map.js
```diff
@@ -0,0 +1,58 @@
...
+      center: this.props.center,
+      zoom: 9,
```
lfender6445: same goes for zoom

.storybook/webpack.config.js
```diff
@@ -28,6 +28,25 @@ module.exports = (storybookBaseConfig, configType) => {
     ]
   })
 
+  const jsConfig = {
```
lfender6445: we can remove this its not being used

stories/theme/Map.css
```diff
@@ -0,0 +1,189 @@
+/* Mapbox styles taken from node_modules/mapbox-gl/dist/mapboxgl.css */
```
tadjohnston: I'd rather use the method I mentioned to @lfender6445 here. I just don't want to get anyone used to using something like this in rent-js

lfender6445: do you have a snippet tad?

packages/react-ui-core/src/Map/Map.js
```diff
@@ -0,0 +1,58 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+import classNames from 'classnames'
+import mapboxgl from 'mapbox-gl/dist/mapbox-gl'
+
+export default class Mapbox extends Component {
```
aribouius: Can this extend `PureComponent` instead?

packages/react-ui-core/src/Map/Map.js
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  static defaultProps = {
+    theme: {},
+  }
+
+  componentDidMount() {
+    mapboxgl.accessToken = this.props.token
+    const map = new mapboxgl.Map({
+      container: 'map',
```
aribouius: Can we make the `container` selector configurable via prop in case we ever need to render multiple maps on a page.

packages/react-ui-core/src/Map/Map.js
```diff
@@ -0,0 +1,58 @@
...
+      <div
+        {...props}
+        style={{}}
+        id="map"
+        className={classNames(
+          className,
+          theme.Map,
+          theme[`Map-${color}`],
+          theme[`Map-${size}`],
+          theme.map
```
aribouius: What is the intended difference between `theme.Map` and `theme.map`?

packages/react-ui-core/src/Map/index.js
```diff
@@ -0,0 +1,2 @@
+export { default } from './Map'
+export { default as Mapbox } from './Map'
```
aribouius: The exported name should match the component, e.g. `export { default as Map } from './Map'`

lfender6445: we could not use Map because it exists on window and will not render via storybook. lets use 

`export { default as Mapbox } from './Mapbox'` and rename the file

stories/theme/Map.css
tadjohnston: something like this:
```
      {
        test: /\.css$/,
        include: [SRC_DIR],
        exclude: [NODE_MODULES_DIR],
        use: ExtractTextPlugin.extract({
          fallback: 'style-loader',
          use: [
            {
              loader: 'css-loader',
              options: {
                modules: true,
                minimize: !env.dev,
                localIdentName: env.dev ? '[path][name]-[local]' : '[hash:base64:5]',
                importLoaders: 1,
              },
            },
            {
              loader: 'postcss-loader',
            },
          ],
        })
      },
      {
        test: /\.css$/,
        include: [NODE_MODULES_DIR],
        exclude: [SRC_DIR],
        use: ExtractTextPlugin.extract({
          fallback: 'style-loader',
          use: ['css-loader', 'postcss-loader']
        })
      }
```

lfender6445:
created react-ui-map please let us know if there is anything else

packages/react-ui-core/package.json
```diff
@@ -50,6 +50,7 @@
     "babel-plugin-react-css-modules": "^3.0.1",
     "babel-runtime": "^6.23.0",
     "classnames": "^2.2.5",
+    "mapbox-gl": "0.39.1",
```
aribouius: Should this still be here?

packages/react-ui-map/src/Mapbox/index.js
```diff
@@ -0,0 +1,56 @@
...
+
+    return (
+      <div
+        style={{}}
+        id="map"
+        className={classNames(
+          className,
+          theme[`Map-${color}`],
+          theme[`Map-${size}`],
+          theme.map
```
aribouius: This should be named `Map`.

packages/react-ui-core/package.json
lfender6445: no. thought i removed ill take care of it.

packages/react-ui-map/src/Mapbox/index.js
```diff
@@ -0,0 +1,56 @@
...
+    container: PropTypes.string,
+  }
+
+  static defaultProps = {
+    theme: {},
+  }
+
+  componentDidMount() {
+    mapboxgl.accessToken = this.props.token
+    const map = new mapboxgl.Map({
```
tcwilson: Does `const map` need to be here? Should we be using the return of this somewhere?

tadjohnston:
Looks great 👍

packages/react-ui-map/src/Mapbox/index.js
lfender6445: eventually the map will be accessible via context https://github.com/rentpath/ag.js/blob/master/src/apps/shared/Map/components/Map.js#L46 but yeah the const is not reused. 

i'll clean this up next iteration

PR #39, opened 2017-08-07
Integrate Map Layers and Sources

maityarkodeb:
[Story](https://www.pivotaltracker.com/story/show/149846374)


packages/react-ui-map/src/Mapbox/index.js
```diff
@@ -28,6 +49,21 @@ export default class Mapbox extends PureComponent {
       zoom: this.props.zoom,
       theme: this.props.theme
     })
+
+    map.on('load', (...args) => {
+      if(this.state.map != {map})
```
tadjohnston: What does this check do here?

maityarkodeb: The global mapbox-gl object (‘map’ here) is saved within the state. This is initialized in the class constructor as null. And then updated within the ComponentDidMount() lifecycle method using setState() method. In case, from any other relevant lifecycle event (for example, especially ComponentWillUpdate() or ComponentDidUpdate()) which causes ComponentDidMount() to be invoked again due to any future code/logic added, to prevent a re-render(), this check has been put. If the ‘map’ is already updated (not null any more), then no need to call setState() again. It will be called only for the first time when state.map is null as initialized in the constructor. So it will go through a render cycle only at that time.
 
Also, all the AJAX calls are done within componentDidMount() as recommended by React. So map.on(‘load’) is handled here and all the source and layers are constructed here for performance reasons.

aribouius: 1. I'm guessing the `{}` around `map` area accidental, either way they shouldn't be there.
2. Always favor using `!==` instead of `!=` (or alternatively `===`).
3. Are you actually using `args` for anything?  If not lets remove it.

packages/react-ui-map/src/Mapbox/index.js
```diff
@@ -28,6 +49,21 @@ export default class Mapbox extends PureComponent {
...
+        this.setState({ map })
+
+      for (let i = 0; i < sources.length; i++) {
```
aribouius: How about using `sources.forEach` instead?

packages/react-ui-map/src/Mapbox/index.js
```diff
@@ -13,13 +21,26 @@ export default class Mapbox extends PureComponent {
     style: PropTypes.string,
     zoom: PropTypes.number,
     container: PropTypes.string,
+    children: PropTypes.object,
```
aribouius: Can this be an array also?

packages/react-ui-map/src/Mapbox/index.js
```diff
@@ -28,6 +49,21 @@ export default class Mapbox extends PureComponent {
       zoom: this.props.zoom,
       theme: this.props.theme
     })
+
+    map.on('load', () => {
+      if(this.state.map !== map)
+        this.setState(map)
```
aribouius: Per our styleguide, please add a whitespace after the `if` and use brackets around the condition body.

PR #41, opened 2017-08-10
refactor map to use child components

maityarkodeb:
Refactoring of components to use child context.

### DIFF
--- a/packages/react-ui-map/src/Mapbox/Layer.js
+++ b/packages/react-ui-map/src/Mapbox/Layer.js
@@ -1,4 +1,4 @@
-import React, { Component } from 'react'
+import { Component } from 'react'
 import PropTypes from 'prop-types'
 
 export default class Layer extends Component {
@@ -8,29 +8,29 @@ export default class Layer extends Component {
     const { map } = this.context
     const { layers } = this.props
 
-    for(let value of layers) {
-        map.addLayer(value.layer)
-    }
+    layers.forEach(element => {
+      map.addLayer(element.layer)
+    })
   }
 
   static propTypes = {
-    layers: PropTypes.array
+    layers: PropTypes.array,
   }
 
   static contextTypes = {
-    map: PropTypes.object
+    map: PropTypes.object,
   }
 
   componentWillUnmount() {
     const { map } = this.context
     const { layers } = this.props
 
-    for(let value of layers) {
-        map.removeLayer(value.layer)
-    }
+    layers.forEach(element => {
+      map.removeLayer(element.layer)
+    })
   }
 
-  render() {
+  render() {  // eslint-disable-line class-methods-use-this
     return null
   }
 }
--- a/packages/react-ui-map/src/Mapbox/Layer.js
+++ b/packages/react-ui-map/src/Mapbox/Layer.js
@@ -1,22 +1,24 @@
 import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 
-export default class MapLayer extends Component {
-  static propTypes = {
-    layers: PropTypes.array
-  }
+export default class Layer extends Component {
+  constructor(props, context) {
+    super(props, context)
 
-  static contextTypes = {
-    map: PropTypes.object
-  }
-
-  componentWillMount() {
     const { map } = this.context
     const { layers } = this.props
 
     for(let value of layers) {
         map.addLayer(value.layer)
     }
+  }
+
+  static propTypes = {
+    layers: PropTypes.array
+  }
+
+  static contextTypes = {
+    map: PropTypes.object
   }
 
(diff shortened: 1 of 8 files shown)


packages/react-ui-map/src/Mapbox/MapLayer.js
```diff
@@ -0,0 +1,34 @@
...
+export default class MapLayer extends Component {
+  static propTypes = {
+    layers: PropTypes.array
+  }
+
+  static contextTypes = {
+    map: PropTypes.object
+  }
+
+  componentWillMount() {
```
tadjohnston: Can you place the code from `componentWillMount` ([from docs](https://facebook.github.io/react/docs/react-component.html#componentwillmount)) in the `constructor` instead?

packages/react-ui-map/src/Mapbox/MapLayer.js
```diff
@@ -0,0 +1,34 @@
...
+    const { map } = this.context
+    const { layers } = this.props
+
+    for(let value of layers) {
+        map.addLayer(value.layer)
```
tadjohnston: Please also ensure `map` is available in order to call a method on it.

packages/react-ui-map/src/Mapbox/MapSource.js
```diff
@@ -0,0 +1,35 @@
...
+  static propTypes = {
+    sources: PropTypes.array,
+    children: PropTypes.node
+  }
+
+  static contextTypes = {
+    map: PropTypes.object
+  }
+
+  componentWillMount() {
```
tadjohnston: Same as the comment above about `componentWillMount`.

packages/react-ui-map/src/Mapbox/MapSource.js
```diff
@@ -0,0 +1,35 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+
+export default class MapSource extends Component {
+  static propTypes = {
+    sources: PropTypes.array,
+    children: PropTypes.node
```
tadjohnston: Where is `children` being used here?

packages/react-ui-map/src/Mapbox/index.js
```diff
@@ -51,14 +47,7 @@ export default class Mapbox extends PureComponent {
...
     map.on('load', () => {
-      if (this.state.map !== map) {
-        this.setState(map)
-      }
-
-      sources.forEach((v, i) => {
-        map.addSource(sources[i].id, sources[i].source)
-        map.addLayer(layers[i].layer)
-      })
+      this.setState({ map })
```
tadjohnston: Does this need to be set on load here or can it be set directly after the instantiation for `new mapboxgl.Map()`?

packages/react-ui-map/src/index.js
```diff
@@ -1,2 +1,4 @@
 export { default } from './Mapbox'
 export { default as Mapbox } from './Mapbox'
+export { default as MapSource } from './Mapbox/MapSource'
```
tadjohnston: I think we should probably change the names to `Source` and `Layer` for the filenames.  This is kinda nitpicky so it's up to what others ideas are.

packages/react-ui-map/src/Mapbox/MapSource.js
maityarkodeb: `children` was not being used.  removed.

packages/react-ui-map/src/Mapbox/MapLayer.js
maityarkodeb: Done.

packages/react-ui-map/src/index.js
maityarkodeb: Changed names to `Source` and `Layer`.

packages/react-ui-map/src/Mapbox/index.js
maityarkodeb: Leaving in on load as per discussion with @tadjohnston.

packages/react-ui-map/src/Mapbox/MapSource.js
maityarkodeb: Addressed.

packages/react-ui-map/src/Mapbox/Layer.js
```diff
@@ -0,0 +1,36 @@
...
+export default class Layer extends Component {
+  constructor(props, context) {
+    super(props, context)
+
+    const { map } = this.context
+    const { layers } = this.props
+
+    for(let value of layers) {
+        map.addLayer(value.layer)
+    }
```
lfender6445: fix indentation for files please, use linter

lfender6445:
👍

PR #42, opened 2017-08-14
Support custom anim prop on <Map />

maityarkodeb:
[Story](https://www.pivotaltracker.com/story/show/150112391)


packages/react-ui-map/src/Mapbox/index.js
```diff
@@ -83,17 +90,20 @@ export default class Mapbox extends PureComponent {
...
-          theme[`Map-${color}`],
-          theme[`Map-${size}`],
-          theme.Map
-        )}
-      >
-        {map && children}
+      <div>
+        <div
+          style={{}}
+          id="map"
```
tadjohnston: `id` should *not* be here.

packages/react-ui-map/src/Mapbox/index.js
```diff
@@ -83,17 +90,20 @@ export default class Mapbox extends PureComponent {
...
-          className,
-          theme[`Map-${color}`],
-          theme[`Map-${size}`],
-          theme.Map
-        )}
-      >
-        {map && children}
+      <div>
+        <div
+          style={{}}
```
tadjohnston: Why is this empty? It shouldn't be here if it's not set.

tadjohnston:
I'm not sure how `packages/react-ui-map/src/Mapbox/index.js` made it through, but `index.js` should really only be used for exporting packages.  Can you please put the mapbox code inside of `packages/react-ui-map/src/Mapbox/Mapbox.js` instead? You can export it through `index.js`.

packages/react-ui-map/src/Mapbox/index.js
maityarkodeb: removed.

packages/react-ui-map/src/Mapbox/index.js
maityarkodeb: changed to use `ref`

packages/react-ui-map/src/Mapbox/Mapbox.js
```diff
@@ -83,17 +89,19 @@ export default class Mapbox extends PureComponent {
...
-          className,
-          theme[`Map-${color}`],
-          theme[`Map-${size}`],
-          theme.Map
-        )}
-      >
-        {map && children}
+      <div>
+        <div
+          ref="map"
```
tadjohnston: This is the incorrect usage of a `ref`. Please refer to the [docs for usage](https://facebook.github.io/react/docs/refs-and-the-dom.html#adding-a-ref-to-a-dom-element)

packages/react-ui-map/src/Mapbox/Mapbox.js
```diff
@@ -83,17 +89,19 @@ export default class Mapbox extends PureComponent {
...
+          className={classNames(
+            className,
+            theme[`Map-${color}`],
+            theme[`Map-${size}`],
+            theme.Map,
+          )}
+        >
+          {map && children}
+        </div>
+        <button className={theme.Button} onClick={this.handleClick}>Fit</button>
```
tadjohnston: Please put these props on their own new lines. Also, why is this button `Fit` in this map file in the first place?  Is it supposed to be here???

packages/react-ui-map/src/Mapbox/Mapbox.js
```diff
@@ -70,6 +70,12 @@ export default class Mapbox extends PureComponent {
       nextState.map !== this.state.map)
   }
 
+  handleClick() {
```
tadjohnston: Is this supposed to be here like this with `fitBounds`?

packages/react-ui-map/src/Mapbox/Mapbox.js
maityarkodeb: working on changes.

packages/react-ui-map/src/Mapbox/Mapbox.js
maityarkodeb: working on moving to storybook

packages/react-ui-map/src/Mapbox/Mapbox.js
maityarkodeb: updated

packages/react-ui-map/src/Mapbox/Mapbox.js
```diff
@@ -56,6 +57,11 @@ export default class Mapbox extends PureComponent {
       nextState.map !== this.state.map)
   }
 
+  componentWillReceiveProps(nextProps) {
+    const { map } = this.state
+    fly(map, nextProps.boundingBox)
```
lfender6445: could we add some defaultProps for boundingBox?

maityarkodeb: I'll add the check for bounding box

tadjohnston:
Looks great! Nice job 👍

lfender6445:
👍

PR #44, opened 2017-08-16
Ratings component

vamshi100:
[Story](https://www.pivotaltracker.com/story/show/149179315)
Rating Stars


packages/react-ui-core/src/Ratings/RatingBar.js
```diff
@@ -0,0 +1,37 @@
...
+        <RatingItem
+          width={`${val * 100}%`}
+          index={currentRating}
+          backgroundColor={backgroundColor}
+          theme={theme}
+        />)
+    }
+    return (
+      <div className="rating-bar">
+      <div className={theme.ratingcomp}>{ratingItems}</div>
```
lfender6445: please indent inner html

packages/react-ui-core/src/Ratings/RatingItem.js
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  render() {
+    const {
+      width,
+      index,
+      backgroundColor,
+      theme,
+    } = this.props
+    const id = `id_${index}`
```
lfender6445: this id is not unique enough

packages/react-ui-core/src/Ratings/RatingBar.js
```diff
@@ -0,0 +1,37 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+import RatingItem from './RatingItem'
+
+export default class RatingBar extends Component {
```
lfender6445: please add some tests for rating bar and item, have a look at the agjs tests for ideas

what do you think about wiring up a snapshot test via jest?  

https://facebook.github.io/jest/docs/snapshot-testing.html

packages/react-ui-core/src/Ratings/RatingItem.js
```diff
@@ -0,0 +1,48 @@
...
+          <linearGradient id={id}>
+            <stop offset="0%" stopOpacity="1" stopColor={backgroundColor}/>
+            <stop offset={width} stopOpacity="1" stopColor={backgroundColor} />
+            <stop offset={width} stopOpacity="0" stopColor={backgroundColor} />
+            <stop offset="100%" stopOpacity="0" stopColor={backgroundColor} />
+          </linearGradient>
+        </defs>
+        <title>Five Pointed Star</title>
+        <path
+          fill={`url(#${id})`}
```
tadjohnston: What is this supposed to be doing?

packages/react-ui-core/src/Ratings/RatingBar.js
```diff
@@ -0,0 +1,37 @@
...
+  static propTypes= {
+    theme: PropTypes.object,
+  }
+  static defaultProps = {
+    theme: {},
+  }
+
+  render() {
+    const { max, rating, backgroundColor, total, theme } = this.props
```
tadjohnston: None of these props are in `propTypes` except for `theme`.  Can we add them please?

packages/react-ui-core/src/Ratings/RatingBar.js
```diff
@@ -0,0 +1,37 @@
...
+          width={`${val * 100}%`}
+          index={currentRating}
+          backgroundColor={backgroundColor}
+          theme={theme}
+        />)
+    }
+    return (
+      <div className="rating-bar">
+        <div className={theme.ratingcomp}>{ratingItems}</div>
+        <div className={theme.label}>(<span>{total}</span> {'Ratings'})</div>
```
tadjohnston: Entering text here should be completely optional. Also, not sure what the `{ }` around `'Ratings'` is doing.

packages/react-ui-core/src/Ratings/RatingBar.js
```diff
@@ -0,0 +1,37 @@
...
+        <RatingItem
+          width={`${val * 100}%`}
+          index={currentRating}
+          backgroundColor={backgroundColor}
+          theme={theme}
+        />)
+    }
+    return (
+      <div className="rating-bar">
+        <div className={theme.ratingcomp}>{ratingItems}</div>
```
tadjohnston: What is the `<div>` wrapper around `ratingItems` needed for?

packages/react-ui-core/src/Ratings/RatingItem.js
```diff
@@ -0,0 +1,48 @@
...
+    theme: {},
+  }
+
+  render() {
+    const {
+      width,
+      index,
+      backgroundColor,
+      theme,
+    } = this.props
```
tadjohnston: Thee props here doesn't exist in `propTypes` with the exception of `theme`. Please add the rest.

packages/react-ui-core/src/Ratings/RatingItem.js
```diff
@@ -0,0 +1,48 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+
+export default class RatingItem extends Component {
+  static propTypes= {
```
tadjohnston: Space needed after `propTypes`

packages/react-ui-core/src/Ratings/RatingItem.js
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  render() {
+    const {
+      width,
+      index,
+      backgroundColor,
+      theme,
+    } = this.props
+    const id = `rItem_${index}`
```
tadjohnston: Why not use a hex instead?

packages/react-ui-core/src/Ratings/RatingItem.js
```diff
@@ -0,0 +1,48 @@
...
+    return (
+      <div className={theme.ratingitem}>
+      <svg
+        xmlns="http://www.w3.org/2000/svg"
+        width="50"
+        height="50"
+        viewBox="0 0 51 48"
+      >
+        <defs>
+          <linearGradient id={id}>
```
tadjohnston: Where is the `id` defined?

packages/react-ui-core/src/Ratings/RatingBar.js
```diff
@@ -0,0 +1,37 @@
...
+    const ratingItems = []
+    let currentRating = rating
+    while (ratingItems.length < max) {
+      const ratingValue = currentRating - 1 > 0 ?
+        1 : currentRating
+      currentRating -= 1
+      const val = ratingValue < 0 ? 0 : ratingValue
+      ratingItems.push(
+        <RatingItem
+          width={`${val * 100}%`}
```
tadjohnston: Can we call this something other than `width`? How about `fillWidth`?

packages/react-ui-core/src/Ratings/RatingBar.js
```diff
@@ -0,0 +1,37 @@
...
+  }
+  static defaultProps = {
+    theme: {},
+  }
+
+  render() {
+    const { max, rating, backgroundColor, total, theme } = this.props
+    const ratingItems = []
+    let currentRating = rating
+    while (ratingItems.length < max) {
```
tadjohnston: What do you think about doing something a little less convoluted like:
```
 get ratingItems() {
    const {
      max,
      rating,
      backgroundColor,
      theme
    } = this.props

    return [...Array(max).keys()].map(index => (
      <RatingItem
        width={`${(rating - index) * 100}%`}
        index={index}
        backgroundColor={backgroundColor}
        theme={theme}
      />
    ))
  }
```

render method would look like this then:
```
<div className={theme.ratingcomp}>{this.ratingItems}</div>
```

packages/react-ui-core/src/Ratings/RatingItem.js
tadjohnston: Why is the `id` here needed?

stories/theme/RatingBar.css
```diff
@@ -0,0 +1,15 @@
+.label {
+  color: #4a90e2;
+  font-size: 50px;
+  text-align: center;
+  display: inline;
+  margin-left: 8px;
+}
+
+.ratingcomp {
+  display:inline-block;
```
tadjohnston: Please add a space after `:` and on `ratingItem` as well.

packages/react-ui-core/src/Ratings/RatingItem.js
```diff
@@ -0,0 +1,48 @@
...
+            <stop offset="0%" stopOpacity="1" stopColor={backgroundColor}/>
+            <stop offset={width} stopOpacity="1" stopColor={backgroundColor} />
+            <stop offset={width} stopOpacity="0" stopColor={backgroundColor} />
+            <stop offset="100%" stopOpacity="0" stopColor={backgroundColor} />
+          </linearGradient>
+        </defs>
+        <title>Five Pointed Star</title>
```
tadjohnston: Can we remove the title please?

packages/react-ui-core/src/Ratings/RatingItem.js
```diff
@@ -0,0 +1,48 @@
...
+    const {
+      width,
+      index,
+      backgroundColor,
+      theme,
+    } = this.props
+    const id = `rItem_${index}`
+    return (
+      <div className={theme.ratingitem}>
+      <svg
```
tadjohnston: Indentation needs to be fixed here.

packages/react-ui-core/src/Ratings/RatingItem.js
vamshi100: color fill is done using offset property inside the linear gradient tag.
the fill over here should contain unique id that matches that of linear gradient because, I'm iterating an array of elements and each should have an unique id else all elements in the array are assuming the properties of first array element, which is undesirable.

tadjohnston: Got it thanks

lfender6445:
Lets pause on this and reflect on some ways we might be able to improve the current setup

- linear gradients are necessary for partial fills (eg star filled 30%)
- this component uses a default SVG of a star, but does not allow design or dev teams to easily change shapes or svgs
- we need to be able to inject linear gradients into an svg

What could we do to faciliate this? Could we create an SVG component inside of react-ui-core that renders children (eg `<svg><linearGradient /><path or polygon data></svg>`)

@aribouius @ninomichaels @tadjohnston

packages/react-ui-core/src/Ratings/RatingItem.js
lfender6445: should we be stripping titles from svgs? aren't they there for accessibility?

tadjohnston:
I agree that using an svg of a star isn't flexible.  I think the svg star should go in another component called `<Star />` outside of the rating item and by default the rating will pass that `<Star />` component down to the `<RatingItem />` as an `item` prop or something.  I think `<RatingItem` should include the logic to linear fill the svgs and include some other props it's missing right now like `strokeWidth`, `strokeColor` and possibly `width` and `height`.

Maybe something more like this (pseudo code):

```
<Star
  backgroundColor // default yellow
  strokeColor // default black
  offset // default 0
  height
  width
/>

<Rating
  item={ <Star />  }
  ratingScale={5} // IMO the name shouldn't be max, but maybe this is wrong too...
/>
```

packages/react-ui-core/src/Ratings/RatingBar.js
vamshi100: <div className={theme.label}>(<span>{total}</span> {'Ratings'})</div>
this line of code was directly copied from ag.js and I didn't think it was necessary to make any modifications to it. But if you feel that {} around Ratings is not necessary, I'll take it out.

tadjohnston: It's not really necessary to use it. Using `{}` in jsx code tells the code to interpret what is inside.  In this case it's just a string so it's an extra step that jsx has to handle.

Please don't rely on ag.js to tell you how to code this.  You should be using ag.js as a guideline, not a definitive answer on how to do implement generic code.  I'm more concerned about enforcing text here in general which should *not* be enforced at all. It should be optional at best. There will be places on the ui where text will not be used and this component needs to be flexible enough to handle that scenario.

packages/react-ui-core/src/Ratings/RatingItem.js
tadjohnston: We can leave it.

packages/react-ui-core/src/Ratings/RatingBar.js
vamshi100: yeah, now that you've mentioned it, even i found it to be redundant. will be fixed in the upcoming commits.

lfender6445:
@tadjohnston The problem with using a Star component is if rentjs ever decided it wants to use a different kind of star or maybe even circles or thumbs, we don't have control over the way the polygon data or path data. 

How we can we gain full control over an svg within react and still remain flexible?

tadjohnston:
@lfender6445 yes we would have control based on my proposal.  The `item` prop would be a component so you could pass any component you choose. The default is the `<Star /> component and that component would be responsible for its own props and possibly logic.  This part can be however you want, so if you want to pass props from the parent to the `item` prop, they should be spread so we don't have to worry about any additional logic / props that need to be added to `<Ratings />` for example.  It's all a matter of where the logic needs to go and what needs to be clearly defined here.

stories/examples/Ratings.js
```diff
@@ -0,0 +1,164 @@
...
+    rating={4.4}
+    total= {'150'}
+    backgroundColor={'royalblue'}
+    theme={RatingBarTheme}
+    viewBox= "0 0 51 48"
+    stroke="black"
+    strokeWidth="2"
+    itemHeight="25px"
+    itemWidth="25px"
+    pathD= "m25,1 6,17h18l-14,11 5,17-15-10-15,10 5-17-14-11h18z"
```
lfender6445: @tadjohnston take another look, with this approach we can control the shape of the svg via props. what do you think?

tadjohnston: Well, to a certain extent yes.  That means 1) you'd always have to have the path data and 2) hope that the svg isn't complex to a point that it needs more than a path.

packages/react-ui-core/src/Ratings/RatingBar.js
```diff
@@ -53,6 +54,7 @@ export default class RatingBar extends Component {
           width={itemWidth}
           pathD={pathD}
           viewBox={viewBox}
+          polygonPoints={polygonPoints}
```
tadjohnston: `polygonPoints` is missing from the `propTypes`. Also, do we really need to enforce all of these props here? Can we spread them instead so we don't have to keep adding support for new props?

packages/react-ui-core/src/Ratings/RatingBar.js
```diff
@@ -0,0 +1,67 @@
...
+    ratingScale: PropTypes.number,
+    rating: PropTypes.number,
+    backgroundColor: PropTypes.string,
+    total: PropTypes.string,
+    stroke: PropTypes.string,
+    strokeWidth: PropTypes.number,
+    itemHeight: PropTypes.string,
+    itemWidth: PropTypes.string,
+    pathD: PropTypes.string.isRequired,
+    viewBox: PropTypes.string.isRequired,
```
aribouius: I'm not sure its a good idea to require svg specific stuff like `pathD` and `viewBox`.  Ideally these concepts should be abstracted away.  In my opinion this component is currently too complicated for general use.

packages/react-ui-core/src/Ratings/RatingBar.js
```diff
@@ -0,0 +1,67 @@
...
+      polygonPoints,
+    } = this.props
+    const ratingItems = []
+    let currentRating = rating
+    while (ratingItems.length < ratingScale) {
+      const ratingValue = currentRating - 1 > 0 ?
+        1 : currentRating
+      currentRating -= 1
+      const val = ratingValue < 0 ? 0 : ratingValue
+      ratingItems.push(
```
aribouius: Please try to use newlines to create better separation between code.  This is currently a nightmare to try to read.

packages/react-ui-core/src/Ratings/RatingItem.js
```diff
@@ -0,0 +1,45 @@
...
+  render() {
+    const {
+      fillWidth,
+      index,
+      backgroundColor,
+      theme,
+      pathD,
+      polygonPoints,
+    } = this.props
+    const id = `rItem_${index}`
```
aribouius: This is falsely assuming that this component will only rendered a single time (per index) on a given page, when it will in all likelihood get rendered multiple times, resulting in numerous duplicate ID's.

packages/react-ui-core/src/Ratings/__tests__/RatingBar-test.js
```diff
@@ -0,0 +1,67 @@
...
+    <RatingBar ratingScale={5}
+    rating={3}
+    total= {'20 Ratings'}
+    backgroundColor={'royalblue'}
+    viewBox= "0 0 51 48"
+    stroke="black"
+    strokeWidth="2"
+    itemHeight="25px"
+    itemWidth="25px"
+    pathD= "m25,1 6,17h18l-14,11 5,17-15-10-15,10 5-17-14-11h18z"
```
aribouius: Please fix indentation here.

packages/react-ui-core/src/Ratings/__tests__/RatingBar-test.js
```diff
@@ -0,0 +1,67 @@
...
+  const tree = renderer.create(
+    <RatingBar ratingScale={0}
+    rating={2.35}
+    backgroundColor={'royalblue'}
+    viewBox= "0 0 51 48"
+    stroke="black"
+    strokeWidth="2"
+    itemHeight="25px"
+    itemWidth="25px"
+    pathD= "m25,1 6,17h18l-14,11 5,17-15-10-15,10 5-17-14-11h18z"/>
```
aribouius: Please fix indentation here.

packages/react-ui-core/src/Ratings/__tests__/RatingBar-test.js
```diff
@@ -0,0 +1,67 @@
...
+    <RatingBar ratingScale={0}
+    rating={4}
+    total= {'20 Ratings'}
+    backgroundColor={'royalblue'}
+    viewBox= "0 0 51 48"
+    stroke="black"
+    strokeWidth="2"
+    itemHeight="25px"
+    itemWidth="25px"
+    pathD= "m25,1 6,17h18l-14,11 5,17-15-10-15,10 5-17-14-11h18z"/>
```
aribouius: Please fix indentation here.

packages/react-ui-core/src/Ratings/__tests__/RatingBar-test.js
```diff
@@ -0,0 +1,67 @@
...
+  ).toJSON()
+  expect(tree).toMatchSnapshot()
+})
+
+it('only renders rating count', () => {
+  const tree = renderer.create(
+    <RatingBar ratingScale={0}
```
aribouius: Please drop the first prop here to a newline.

packages/react-ui-core/src/Ratings/__tests__/RatingBar-test.js
```diff
@@ -0,0 +1,67 @@
...
+    itemWidth="25px"
+    pathD= "m25,1 6,17h18l-14,11 5,17-15-10-15,10 5-17-14-11h18z"
+    />
+  ).toJSON()
+  expect(tree).toMatchSnapshot()
+})
+
+it('does not render anything when rating scale is 0 no matter what the rating might be', () => {
+  const tree = renderer.create(
+    <RatingBar ratingScale={0}
```
aribouius: Please drop the first prop here to a newline.

packages/react-ui-core/src/Ratings/__tests__/RatingBar-test.js
```diff
@@ -0,0 +1,67 @@
+import React from 'react'
+import renderer from 'react-test-renderer'
+import RatingBar from '../RatingBar'
+
+it('renders correctly', () => {
+  const tree = renderer.create(
+    <RatingBar ratingScale={5}
```
aribouius: Please drop the first prop here to a newline.

packages/react-ui-core/src/Ratings/__tests__/RatingBar-test.js
```diff
@@ -0,0 +1,67 @@
...
+    itemHeight="25px"
+    itemWidth="25px"
+    pathD= "m25,1 6,17h18l-14,11 5,17-15-10-15,10 5-17-14-11h18z"/>
+  ).toJSON()
+  expect(tree).toMatchSnapshot()
+})
+
+it('renders partial ratings', () => {
+  const tree = renderer.create(
+    <RatingBar ratingScale={5}
```
aribouius: Please drop the first prop here to a newline.

packages/react-ui-core/src/Ratings/__tests__/RatingBar-test.js
```diff
@@ -0,0 +1,67 @@
...
+    rating={2.35}
+    total= {'20 Ratings'}
+    backgroundColor={'royalblue'}
+    viewBox= "0 0 51 48"
+    stroke="black"
+    strokeWidth="2"
+    itemHeight="25px"
+    itemWidth="25px"
+    pathD= "m25,1 6,17h18l-14,11 5,17-15-10-15,10 5-17-14-11h18z"/>
```
aribouius: Please fix indentation here.

packages/react-ui-core/src/Ratings/RatingBar.js
```diff
@@ -0,0 +1,49 @@
...
+      ratingScale,
+      rating,
+      RatingItem,
+      ...props
+    } = this.props
+
+    return [...Array(ratingScale).keys()].map(index => (
+      <RatingItem
+        key={randomId('rating')}
+        id={randomId('rating-item')}
```
aribouius: The purpose of the `key` prop is to help `React` determine whether the component needs to be re-rendered/inserted into the DOM.  By generating a random key every time this component renders you are blocking a potential performance enhancement.  I think it would probably be suitable to just use the `index` value here.

Also, instead of generating a random `id` here every time this renders, lets move that responsibility into the `Star` component, and have it generate it either in the constructor, or create it _lazily_ like [here](https://github.com/rentpath/react-ui/blob/master/packages/react-ui-core/src/Form/Field.js#L118).

packages/react-ui-core/src/Ratings/SampleSvgs/Circle.js
```diff
@@ -0,0 +1,51 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+import { randomId } from '@rentpath/react-ui-utils'
+import LinearGradient from '../LinearGradient'
+
+
```
aribouius: Please remove extra newline here.

packages/react-ui-core/src/Ratings/SampleSvgs/Square.js
```diff
@@ -0,0 +1,53 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+import { randomId } from '@rentpath/react-ui-utils'
+import LinearGradient from '../LinearGradient'
+
+
```
aribouius: Please remove newline here.

packages/react-ui-core/src/Ratings/SampleSvgs/Star.js
```diff
@@ -0,0 +1,53 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+import { randomId } from '@rentpath/react-ui-utils'
+import LinearGradient from '../LinearGradient'
+
+
```
aribouius: Please remove newline here.

packages/react-ui-core/src/Ratings/SampleSvgs/Star.js
```diff
@@ -0,0 +1,53 @@
...
+            backgroundColor={backgroundColor}
+          />
+          <path
+            fill={`url(#${id})`}
+            stroke="#000"
+            strokeWidth="3"
+            d="m25,1 6,17h18l-14,11 5,17-15-10-15,10 5-17-14-11h18z"
+          />
+        </svg>
+      </div>
```
aribouius: Do we really need the wrapping `div`?

stories/examples/Ratings.js
```diff
@@ -0,0 +1,106 @@
+import React from 'react'
+import { RatingBar, Star, Square, Circle }  from 'react-ui-core/src'
+import { RatingBarTheme } from '../theme'
+
+
```
aribouius: Please remove newline.

packages/react-ui-core/src/Ratings/RatingBar.js
tadjohnston: `index` cannot be used here due to the `LinearGradient` setting this as `id` and the `id` being referenced by the `fill={`url(#${id})`}` in the `Star`. You have a chance for id collisions if we do this. Maybe we can reference differently in the `fill` to fix the issue.

packages/react-ui-core/src/Ratings/SampleSvgs/Circle.js
```diff
@@ -0,0 +1,47 @@
...
+          width="50"
+          height="50"
+          viewBox="0 0 600 300"
+        >
+          <LinearGradient
+            id={id}
+            width={width}
+            backgroundColor={backgroundColor}
+          />
+          <circle cx="250" cy="150" r="145" fill="url(#lg)"
```
tadjohnston: Please fix formatting for this circle;  indentation needs adjusted and  props should be on their own lines.

packages/react-ui-core/src/Ratings/SampleSvgs/Square.js
```diff
@@ -0,0 +1,49 @@
...
+    theme: PropTypes.object,
+    backgroundColor: PropTypes.string,
+    stroke: PropTypes.string,
+    strokeWidth: PropTypes.number,
+    width: PropTypes.string,
+  }
+
+  static defaultProps = {
+    theme: {},
+    width: '0%',
```
tadjohnston: Im pretty sure you can just use `0` here.

packages/react-ui-core/src/Ratings/SampleSvgs/Circle.js
```diff
@@ -0,0 +1,47 @@
+import React, { Component } from 'react'
```
tadjohnston: There is no reason to name this directory `SampleSvgs`.  Please either move `Circle.js` out or rename the directory accordingly.

packages/react-ui-core/src/Ratings/index.js
```diff
@@ -0,0 +1,4 @@
+export { default as RatingBar } from './RatingBar'
+export { default as Star } from './SampleSvgs/Star'
+export { default as Square } from './SampleSvgs/Square'
```
tadjohnston: Square and Circle do *not* belong in `src` code. They should only be included in storybook.

packages/react-ui-core/src/Ratings/LinearGradient.js
```diff
@@ -0,0 +1,35 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+import { randomId } from '@rentpath/react-ui-utils'
+
+export default class LinearGradient extends Component {
+  static propTypes = {
+    id: PropTypes.string,
+    fillWidth: PropTypes.string,
+    backgroundColor: PropTypes.string,
```
tadjohnston: Maybe we should call this just color?

packages/react-ui-core/src/Ratings/RatingBar.js
aribouius: I meant to use `index` for the `key`.

tadjohnston: Got it 👍

aribouius: What are your thoughts on moving the `id` generation to the `Star` component though?

tadjohnston: The reason I put that here, is so if we ever create another component outside of react-ui instead in place of the `<Star />`, it doesn't have to do the logic again.  That was the only reason.

packages/react-ui-core/src/Ratings/RatingBar.js
```diff
@@ -0,0 +1,49 @@
...
+
+  render() {
+    const {
+      total,
+      theme,
+    } = this.props
+    return (
+      <div className={theme.ratingbar}>
+        <div className={theme.ratingcomp}>{this.ratingItems}</div>
+        <div className={theme.label}>{total}</div>
```
tadjohnston: If there is no `total` passed in, we shouldn't be using this line.  Can we make this optional please?

packages/react-ui-core/src/Ratings/Star.js
```diff
@@ -0,0 +1,49 @@
...
+
+  render() {
+    const {
+      id,
+      Color,
+      width,
+    } = this.props
+
+    return (
+        <svg
```
tadjohnston: Indentation needs to be fixed here.

stories/SampleRatingComp/Circle.js
```diff
@@ -0,0 +1,52 @@
...
+
+  render() {
+    console.log('props',this.props)
+    const {
+      id,
+      Color,
+      width,
+    } = this.props
+
+    return (
```
tadjohnston: Indentation needs to be fixed here.

stories/SampleRatingComp/Circle.js
```diff
@@ -0,0 +1,52 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+import { randomId } from 'react-ui-utils/src'
+import { LinearGradient } from 'react-ui-core/src'
+
+export default class Circle extends Component {
+  static propTypes = {
+    id: PropTypes.string,
+    theme: PropTypes.object,
+    Color: PropTypes.string,
```
tadjohnston: We only capitalize if the prop type is a component.

stories/SampleRatingComp/Square.js
```diff
@@ -0,0 +1,49 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+import { randomId } from 'react-ui-utils/src'
+import { LinearGradient } from 'react-ui-core/src'
+
+export default class Square extends Component {
+  static propTypes = {
+    id: PropTypes.string,
+    theme: PropTypes.object,
+    Color: PropTypes.string,
```
tadjohnston: We only capitalize if the prop type is a component.

stories/SampleRatingComp/Square.js
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  render() {
+    const {
+      id,
+      Color,
+      width,
+    } = this.props
+
+    return (
```
tadjohnston: Indentation needs to be fixed.

packages/react-ui-core/src/Ratings/LinearGradient.js
```diff
@@ -0,0 +1,35 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+import { randomId } from '@rentpath/react-ui-utils'
+
+export default class LinearGradient extends Component {
+  static propTypes = {
+    id: PropTypes.string,
+    fillWidth: PropTypes.string,
+    Color: PropTypes.string,
```
tadjohnston: We only capitalize if the prop type is a component.

packages/react-ui-core/src/Ratings/RatingBar.js
```diff
@@ -0,0 +1,49 @@
...
+    ratingScale: PropTypes.number,
+    rating: PropTypes.number,
+    total: PropTypes.string,
+    RatingItem: PropTypes.node,
+  }
+  static defaultProps = {
+    theme: {},
+    RatingItem: Star,
+  }
+// <div className={theme.label}>{total}</div>
```
tadjohnston: This needs to be removed.

packages/react-ui-core/src/Ratings/index.js
```diff
@@ -0,0 +1,3 @@
+export { default as RatingBar } from './RatingBar'
```
tadjohnston: Can we also add the following export to match the way the other components are done?
```
export { default } from './RatingBar'
```

stories/SampleRatingComp/Circle.js
```diff
@@ -0,0 +1,52 @@
...
+    width: PropTypes.string,
+  }
+
+  static defaultProps = {
+    theme: {},
+    width: '0',
+  }
+
+  render() {
+    console.log('props',this.props)
```
tadjohnston: Whoops!

stories/examples/Ratings.js
```diff
@@ -0,0 +1,106 @@
+import React from 'react'
+import { RatingBar }  from 'react-ui-core/src'
+import { Square, Circle } from '../SampleRatingComp'
+import { RatingBarTheme } from '../theme'
+
+export const DefaultRatings = (
+  <RatingBar
+    ratingScale={5}
+    rating={2.35}
+    // total= {'20 Ratings'}
```
tadjohnston: This comment shouldn't be here.

stories/SampleRatingComp/Circle.js
```diff
@@ -0,0 +1,52 @@
+import React, { Component } from 'react'
```
tadjohnston: Please rename the directory to `SampleRatingItem`

stories/examples/Ratings.js
```diff
@@ -0,0 +1,106 @@
...
+
+export const Partial3 = (
+  <RatingBar
+    ratingScale={5}
+    rating={0.75}
+    total= {'150'}
+    Color={'royalblue'}
+    theme={RatingBarTheme}
+  />
+)
```
tadjohnston: You need examples that use both the `Circle` and `Square` components.

stories/examples/Ratings.js
```diff
@@ -0,0 +1,106 @@
...
+import { RatingBar }  from 'react-ui-core/src'
+import { Square, Circle } from '../SampleRatingComp'
+import { RatingBarTheme } from '../theme'
+
+export const DefaultRatings = (
+  <RatingBar
+    ratingScale={5}
+    rating={2.35}
+    // total= {'20 Ratings'}
+    Color={'royalblue'}
```
tadjohnston: Please change these colors to just a string:

```
color="royalBlue"
```

The rest below need changed as well.

packages/react-ui-core/src/Ratings/RatingBar.js
aribouius: That sort of makes sense, but the `id` is really only used internally by the `Star` component to link the svg nodes, so imo this parent component has no need to know about it.

packages/react-ui-core/src/Ratings/Star.js
```diff
@@ -0,0 +1,48 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+import LinearGradient from './LinearGradient'
+
+export default class Star extends Component {
+  static propTypes = {
+    id: PropTypes.string,
```
aribouius: If we are going to keep having the parent pass in this `id` prop it needs to be required.

stories/SampleRatingItem/Circle.js
```diff
@@ -0,0 +1,51 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+import { randomId } from 'react-ui-utils/src'
+import { LinearGradient } from 'react-ui-core/src'
+
+export default class Circle extends Component {
+  static propTypes = {
+    id: PropTypes.string,
```
aribouius: Same story here.

stories/SampleRatingItem/Square.js
```diff
@@ -0,0 +1,49 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+import { randomId } from 'react-ui-utils/src'
+import { LinearGradient } from 'react-ui-core/src'
+
+export default class Square extends Component {
+  static propTypes = {
+    id: PropTypes.string,
```
aribouius: And here.

packages/react-ui-core/src/Ratings/RatingBar.js
```diff
@@ -0,0 +1,49 @@
...
+import { randomId } from '@rentpath/react-ui-utils'
+import Star from './Star'
+
+export default class RatingBar extends Component {
+  static propTypes = {
+    theme: PropTypes.object,
+    ratingScale: PropTypes.number,
+    rating: PropTypes.number,
+    total: PropTypes.string,
+    RatingItem: PropTypes.node,
```
lfender6445: please change this to PropTypes.func then i am 👍

PR #46, opened 2017-08-17
Add lib to publish

lfender6445:
I goofed and the lib folder didn't get published last time.

### DIFF
--- a/packages/react-ui-map/package.json
+++ b/packages/react-ui-map/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@rentpath/react-ui-map",
-  "version": "0.0.03",
+  "version": "0.0.04",
   "main": "lib/index.js",
   "license": "MIT",
   "scripts": {


tcwilson:
:+1:

packages/react-ui-map/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@rentpath/react-ui-map",
-  "version": "0.0.03",
+  "version": "0.0.04",
```
aribouius: Whats the deal with the `.0x`?  It should just be `0.0.4`.

lfender6445: hmm. people keep telling me different things or maybe i just don't understand our versioning setup. 

if you want me to change it i will. what rules should i be following for versioning so i can ensure i'm doing it the right way from now on

aribouius: I'm not sure where you heard you should do this, but its not necessary.  Its a simple zero based counter (for each type) - you don't need to add any leading `0`'s.

aribouius: To clarify - this _does not_ use a decimal system.

lfender6445: going to leave it as is for now since its already been published well do. in that case, i'm assuming the next published version should be `0.0.1`?

aribouius: Lol no I think that might be confusing, lets just go with `0.0.5`.

PR #48, opened 2017-08-25
fixing all eslint errors

tadjohnston:
No Story

Fixing lint errors after upgraded to `@rentpath/eslint-config-rentpath`


lfender6445:
please lint fix the stories folder + update the path in package.json to include this as well

👍

tadjohnston:
Doh! Forgot to include stories. This is going to take a bit and lerna currently only runs on `packages`.

aribouius:
👍

PR #49, opened 2017-08-28
fix mapbox unmounting errors

maityarkodeb:
- fixes errors when trying to remove a source or layer that no longer exists in the map's style
- ~~linting of the map tests is unrelated to PR but saw that it needed fix~~ (fixed by Tad's PR)


packages/react-ui-map/src/Mapbox/Source.js
```diff
@@ -24,16 +24,6 @@ export default class Source extends Component {
...
-    const { map } = this.context
-    const { id, type, data } = this.props
-
-    map.removeSource(id, {
-      type,
-      data,
-    })
-  }
-
   render() {  // eslint-disable-line class-methods-use-this
```
tadjohnston: This `// eslint-disable-line class-methods-use-this`  line has already been removed.

lfender6445:
what are some of the scenarios that will cause a layers removal before an umount?

maityarkodeb:
@lfender6445 I don't think any scenario like this exists.  As soon as you navigate to another page, the map object is destroyed which triggers the error.  I was initially doing the layer removal as a bit of cleanup to free up memory, but was working on a single page and so didn't see the errors.  I think getting rid of the code is fine as it still works without the errors.

lfender6445:
got it. i'm 👍  please rebase and lint

lfender6445:
oh and bump version

tadjohnston:
@maityarkodeb what if you're using a single page application? Does it still get destroyed?  You might need to check that.

maityarkodeb:
@tadjohnston Good call.  I didn't account for that scenario.  Instead I have implemented code to safely remove the layer by guarding the call as suggested here:  https://github.com/mapbox/mapbox-gl-js/issues/4466

I am also doing a similar thing as ag.js for accounting for sources by checking for an existing source via `map.getSource()`.  These two changes remove the errors.

packages/react-ui-map/src/Mapbox/Layer.js
```diff
@@ -23,7 +23,9 @@ export default class Layer extends Component {
     const { map } = this.context
     const { layer } = this.props
 
-    map.removeLayer(layer)
+    if (map.getLayer(layer)) {
```
tadjohnston: Just to be on the safe side, I think we should add a safeguard like so:
```
if (map && map.getLayer(layer)) {
```

might be overkill, but at least we don't throw an error it for some reason map doesn't exist. Same below.

packages/react-ui-map/src/Mapbox/Source.js
```diff
@@ -24,16 +27,6 @@ export default class Source extends Component {
     map: PropTypes.object,
   }
 
-  componentWillUnmount() {
```
tadjohnston: Do we not still want this?

maityarkodeb: Layers need to be removed before sources or else there will be an error.  I thought it made more sense to go the ag.js route and only add the source in the constructor if it doesn't already exist, instead of trying to remove it explicitly.

tadjohnston: Ok, sounds like a good approach.

packages/react-ui-map/src/Mapbox/Layer.js
maityarkodeb: Makes sense.  Added.

tadjohnston:
You'll need to rebase before moving forward.

PR #53, opened 2017-09-06
Allow updates to source data

lfender6445:
Allow updates to source data

- remove fitbounds wrapper

### DIFF
--- a/packages/react-ui-map/src/Mapbox/Mapbox.js
+++ b/packages/react-ui-map/src/Mapbox/Mapbox.js
@@ -10,7 +10,10 @@ export default class Mapbox extends PureComponent {
     color: PropTypes.string,
     theme: PropTypes.object,
     token: PropTypes.string,
-    center: PropTypes.array,
+    center: PropTypes.oneOfType([
+      PropTypes.object,
+      PropTypes.array,
+    ]),
     flyOnCenterChange: PropTypes.bool,
     style: PropTypes.string,
     zoom: PropTypes.number,
@@ -42,18 +45,20 @@ export default class Mapbox extends PureComponent {
   }
 
   componentDidMount() {
-    if (this.props.center) {
-      const map = new mapboxgl.Map({
-        container: this.container,
-        style: this.props.style,
-        center: this.props.center,
-        zoom: this.props.zoom,
-        theme: this.props.theme,
-      })
-      map.on('load', () => {
-        this.setState({ map })
-      })
+    const opts = {
+      container: this.container,
+      style: this.props.style,
+      center: this.props.center,
+      zoom: this.props.zoom,
+      theme: this.props.theme,
     }
+
+    if (!opts.center) delete opts.center
+
+    const map = new mapboxgl.Map(opts)
+    map.on('load', () => {
+      this.setState({ map })
+    })
   }
 
   componentWillReceiveProps(nextProps) {
--- a/packages/react-ui-map/src/Mapbox/Mapbox.js
+++ b/packages/react-ui-map/src/Mapbox/Mapbox.js
@@ -2,7 +2,6 @@ import React, { PureComponent } from 'react'
 import PropTypes from 'prop-types'
 import classNames from 'classnames'
 import mapboxgl from 'mapbox-gl/dist/mapbox-gl'
-import fly from './fly'
 
 export default class Mapbox extends PureComponent {
   static propTypes = {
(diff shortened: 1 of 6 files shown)


packages/react-ui-map/src/Mapbox/Mapbox.js
```diff
@@ -28,36 +26,39 @@ export default class Mapbox extends PureComponent {
...
-      style: this.props.style,
-      center: this.props.center,
-      zoom: this.props.zoom,
-      theme: this.props.theme,
-    })
-
-    map.on('load', () => {
-      this.setState({ map })
-    })
+    if (this.props.center) {
```
tadjohnston: So if there is no `center` prop we don't render the map?

packages/react-ui-map/src/Mapbox/Mapbox.js
```diff
@@ -28,36 +26,39 @@ export default class Mapbox extends PureComponent {
...
+      })
+    }
   }
 
   componentWillReceiveProps(nextProps) {
     const { map } = this.state
 
-    if (nextProps.boundingBox) {
-      fly(map, nextProps.boundingBox)
+    if (map && nextProps.center) {
```
tadjohnston: Don't you want to check if current `center` prop is different from the next `center` prop?  I personally think this is a little much considering this component is supposed to be generic. I don't think we should be enforcing a `flyTo` at all.  Maybe adding a prop that toggles whether or not we do this?
```
flyOnCenterChange: false
``` 

@aribouius thoughts?

aribouius: Luke please explain the rationale for this change.

lfender6445: we we don't get a bounding box from any of our data endpoints. we just get a center or centroid for the city. 

if you really want to center in on a bounding box you just call `map.fitbounds(bbox)` which is the same as saying `map.setCenter(bbox.getCenter())` with some adjustments for zooming and padding.

packages/react-ui-map/src/Mapbox/Mapbox.js
lfender6445: so i'm confused, why render the map if we don't have a location?

lfender6445: also it breaks if no center is provided.

packages/react-ui-map/src/Mapbox/Mapbox.js
lfender6445: for our usecase, it works like this:

render map with center -> get results -> fitbounds on those results

then

user changes location -> get new results -> fit bounds on those results. 

i like the idea of checking the center change and toggling fly behavior.

packages/react-ui-map/src/Mapbox/Mapbox.js
tadjohnston: Why can't we just have a default prop for `center`? If you don't pass it in, you get the default e.g. Atlanta.

tadjohnston: From the [docs](https://www.mapbox.com/mapbox-gl-js/api/#map)
> The inital geographical centerpoint of the map. If  center is not specified in the constructor options, Mapbox GL JS will look for it in the map's style object. If it is not specified in the style, either, it will default to  [0, 0] Note: Mapbox GL uses longitude, latitude coordinate order (as opposed to latitude, longitude) to match GeoJSON.

lfender6445:
comments addreseed

packages/react-ui-map/src/Mapbox/Mapbox.js
```diff
@@ -24,30 +26,36 @@ export default class Mapbox extends PureComponent {
...
+    const opts = {
       container: this.container,
       style: this.props.style,
       center: this.props.center,
       zoom: this.props.zoom,
       theme: this.props.theme,
-    })
+    }
 
+    if (!opts.center) delete opts.center
```
aribouius: This feels odd.  How about adding `center` to `opts` if it has a value.

aribouius:
Other than my last comment this is 👍

packages/react-ui-map/src/Mapbox/Mapbox.js
lfender6445: good idea

PR #55, opened 2017-09-14
Pass mapbox-gl library to children via React context

tcwilson:
[Story](https://www.pivotaltracker.com/story/show/151062514)

Updating this comment.

I talked to @aribouius earlier. We couldn't get dynamic import to work right now. I've shimmed the `Mapbox` component in `rent-js` to only load this library when on the client side. The `Mapbox` component here will be responsible for loading `mapbox-gl` and passing it down via the `MapboxGL` context variable.

~~Since the component is called `Mapbox` and the library is already setup here, I figured it would be easier to just import the `mapbox-gl` here and have it provide access to the library via React context. Hopefully if we ever need a map again, we won't have to deal with this issue again.~~

~~@tadjohnston, @aribouius -- while this builds without error for both dev/prod mode, it does not seem to create a new chunk in webpack. Actually, it looks like it is creating a new chunk. But SSR is still broken. Any ideas as to what might be going on here? I've tried adding `dynamic-import-node` and `syntax-dynamic-import` to the `.babelrc` in `rent-js`, but still nothing.~~

### DIFF
--- a/packages/react-ui-map/src/Mapbox/Mapbox.js
+++ b/packages/react-ui-map/src/Mapbox/Mapbox.js
@@ -1,7 +1,7 @@
 import React, { PureComponent } from 'react'
 import PropTypes from 'prop-types'
 import classNames from 'classnames'
-import mapboxgl from 'mapbox-gl/dist/mapbox-gl'
+import MapboxGL from 'mapbox-gl/dist/mapbox-gl'
 
 export default class Mapbox extends PureComponent {
   static propTypes = {
@@ -22,6 +22,7 @@ export default class Mapbox extends PureComponent {
 
   static childContextTypes = {
     map: PropTypes.object,
+    MapboxGL: PropTypes.object,
   }
 
   static defaultProps = {
@@ -31,7 +32,6 @@ export default class Mapbox extends PureComponent {
 
   constructor(props, context) {
     super(props, context)
-    mapboxgl.accessToken = props.token
 
     this.state = {
       map: null,
@@ -41,20 +41,12 @@ export default class Mapbox extends PureComponent {
   getChildContext() {
     return {
       map: this.state.map,
+      MapboxGL,
     }
   }
 
   componentDidMount() {
-    const opts = {
-      container: this.container,
-      style: this.props.style,
-      zoom: this.props.zoom,
-      theme: this.props.theme,
-    }
-
-    if (this.props.center) opts.center = this.props.center
-
-    const map = new mapboxgl.Map(opts)
+    const map = this.setupMapbox()
     map.on('load', () => {
       this.setState({ map })
     })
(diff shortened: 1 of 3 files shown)


packages/react-ui-map/src/Mapbox/Mapbox.js
```diff
@@ -79,6 +74,18 @@ export default class Mapbox extends PureComponent {
...
 
+  setupMapbox() {
+    this.MapboxGL.accessToken = this.props.token
+    const { center, style, zoom, theme } = this.props
+    return new this.MapboxGL.Map({
+      container: this.container,
+      style,
+      zoom,
+      theme,
+      ...(center ? { center } : {}),
```
tadjohnston: 👍

tadjohnston:
Btw, just found this. Maybe we can use it: https://github.com/mapbox/mapbox-gl-js-mock

packages/react-ui-map/src/Mapbox/Mapbox.js
```diff
@@ -41,20 +40,16 @@ export default class Mapbox extends PureComponent {
...
-    const opts = {
-      container: this.container,
-      style: this.props.style,
-      zoom: this.props.zoom,
-      theme: this.props.theme,
-    }
+  async componentDidMount() {
+    /* Load mapbox-gl on client only.
+       Attempting to load mapbox-gl on server will break server-side rendering. */
+    this.MapboxGL = await import('mapbox-gl/dist/mapbox-gl')
```
tadjohnston: You might also need a wrapper around this.  Say this component unmounted.  I might assume this would load in mapbox again and definitely run `this.setupMapbox()` again.  Something like this?
```
if (window && !window.mapboxgl) {
  // import
}
```

You could also use this anywhere else in code.  It will still work for both SSR and client.

Not sure why it's not available on rent-js right now, but through the [examples](https://www.mapbox.com/mapbox-gl-js/examples/) it is and it's something I've used before.

screen shot 2017-09-14 at 7 29 19 am

Also, there's also this one which we _really_ should be using as well.  I think we used it on rent in flightjs

https://github.com/mapbox/mapbox-gl-supported

tcwilson: I'll have to check. I would assume that `import()` would know whether something is loaded/cached, or at the very least the browser would still have the JS cached. It should be easy to check though, just switch from Map view to List view and back. But I think we need to call `setupMapbox()` when remounted anyway, to get an instance of the Map

As for the `mapbox-gl-supported`, not really sure what we would do there. We have nothing to fall back to so the user is out of luck I think.

tadjohnston: yep, agreed on the call of `setupMapbox()`.

tcwilson:
@tadjohnston -- The PR comment at the top has been updated. Want to review this again?

For the matching `rent-js` PR, see rentpath/rent-js#118

lfender6445:
don't forget to update https://github.com/rentpath/rent-js/blob/master/src/app/components/small/View/FitBounds.js#L3 when you patch rent js please

lfender6445:
oh nvm you already did

lfender6445:
👍

lfender6445:
go ahead and bump version in package.json ill build and publish once its in

packages/react-ui-map/src/Mapbox/__tests__/Mapbox-test.js
```diff
@@ -0,0 +1,41 @@
+import React from 'react'
+import { mount } from 'enzyme'
+import Mapbox from '../Mapbox'
+
+jest.mock('mapbox-gl/dist/mapbox-gl', () => ({
```
tadjohnston: Nice! 👍

PR #56, opened 2017-09-15
fixing bug with center comparison

tadjohnston:
No Story

When the center object ins't changed, it's still creating a new object so the equality check here is invalid every time.  In order to alleviate this issue, we're not checking the actual `lat` and `lng` instead.


lfender6445:
👍

maityarkodeb:
👍

PR #57, opened 2017-09-16
Fix React cannot find id Invariant Error

lfender6445:
Reserve container for injected map canvas only, react components must render outside of this


tcwilson:
👍

PR #58, opened 2017-09-18
Optional button on collapse component

lfender6445:
- input range now also accepts object or string for formatting header


lfender6445:
i need this pubished for board review

PR #59, opened 2017-09-19
Add react-ui-tracking

aribouius:
- Introduces a new child repo `react-ui-tracking` for components related to event tracking.
- Adds components taken from AG.js (`Tracker`, `Tagging`, `PageView`) which have been modified to be `redux` agnostic while still supporting similar functionality.


packages/react-ui-tracking/src/Tracker.js
```diff
@@ -0,0 +1,87 @@
...
+      this._pageView = { ...props }
+      this.update().then(() => {
+        this.props.pageView(this._pageView)
+        this._pageView = null
+      })
+    }
+  }
+
+  render() {
+    return Array.isArray(this.props.children) ? (
```
lfender6445: can you elaborate on this piece? why wrap with a div?

aribouius: because you can't render an array of children (yet).

lfender6445:
I'm 👍  and we can extend as needed but mind adding a readme? Whats left?

ninomichaels:
LGTM 👍 , agree with Luke though, maybe some basic usage examples?

aribouius:
@lfender6445 @lfender6445 thanks guys I need to get this out so I can push a separate PR on `rent-js`, but I'll add some docs today.

PR #62, opened 2017-09-27
Remove map state

lfender6445:
Destroy the entire map on unmount

### DIFF
--- a/packages/react-ui-map/src/Mapbox/Mapbox.js
+++ b/packages/react-ui-map/src/Mapbox/Mapbox.js
@@ -65,9 +65,8 @@ export default class Mapbox extends PureComponent {
     }
   }
 
-  shouldComponentUpdate(nextProps, nextState) {
-    return (nextProps.children !== this.props.children ||
-      nextState.map !== this.map)
+  shouldComponentUpdate(nextProps) {
+    return nextProps.children !== this.props.children
   }
 
   componentWillUnmount() {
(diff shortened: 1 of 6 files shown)


packages/react-ui-map/src/Mapbox/Mapbox.js
```diff
@@ -67,7 +67,11 @@ export default class Mapbox extends PureComponent {
 
   shouldComponentUpdate(nextProps, nextState) {
     return (nextProps.children !== this.props.children ||
-      nextState.map !== this.state.map)
+      nextState.map !== this.map)
```
tcwilson: this condition, `nextState.map !== this.map` no longer makes sense. As `map` isn't part of state, and `this.map` should get destroyed on component unmount anyway

paulpostel:
I'm not really sure what the issue was that led to this.  This seems to mainly delay initialization to `ComponentDidMount`  the PR description says `Destroy the entire map on unmount` is that happening?

lfender6445:
the PR description says Destroy the entire map on unmount is that happening?

yes at `componentWillUnmount -> map.remove()`

There was actually 2 issues - 

  - map was not loading reliably or consistently (hence the map on styleload change)
  - map resources were not being cleaned up properly between unmounts, so rather than worry about any of them, we destroy the entire map instead.

packages/react-ui-map/src/Mapbox/Mapbox.js
```diff
@@ -111,7 +113,7 @@ export default class Mapbox extends PureComponent {
             theme.Map,
           )}
         />
-        {map && children}
+        {(this.state.loaded && this.map) && children}
```
tadjohnston: You can probably remove the parens.

paulpostel:
👍

PR #65, opened 2017-10-09
Disable Map Rotation

pkanuganti:
Story : https://www.pivotaltracker.com/n/projects/1946211/stories/151559598

### DIFF
--- a/packages/react-ui-map/package.json
+++ b/packages/react-ui-map/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@rentpath/react-ui-map",
-  "version": "0.0.14",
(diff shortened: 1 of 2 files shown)


packages/react-ui-map/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@rentpath/react-ui-map",
-  "version": "0.0.13",
+  "version": "0.0.14",
```
tadjohnston: Technically this is a minor version since you're introducing new features that are backwards compatible.  Please make this `0.1.0`

lfender6445:
👍

PR #66, opened 2017-10-10
upgrading packages and react 16

tadjohnston:
[Story](https://rentpath.leankit.com/card/558597830)

* Upgrades to all packages including react 16. 
* Fixed ability to run all tests for all packages from the root directory.
* Added `.npmignore` to all packages so it doesn't include src and such.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,12 @@
+{
+  "presets": [
+    "es2015",
+    "react"
+  ],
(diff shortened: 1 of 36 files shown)


ninomichaels:
LGTM 👍  tests and lint ran fine from root!

tcwilson:
looks ok to me 👍

PR #68, opened 2017-10-16
fixes client side map issues

tadjohnston:
[Story](https://rentpath.leankit.com/card/558597830)

- fixes issues with client side vs ssr
- adds require ensure so webpack can chunk split it separately into its
own bundle

NOTE: Reverted mapbox-gl bump due to warning message about css missing. Will need to address in another PR.


ninomichaels:
👏

tcwilson:
LGTM 👍

PR #69, opened 2017-10-19
fixes ensure

tadjohnston:


orionstein:
Are we sure this is correct? Shouldn't it also be 

require.ensure(['mapbox-gl/dist/mapbox-gl'], map => {
      this.MapboxGL = map
      this.setupMapbox()
   })

Also, regarding the fallback, I don't think it helps the way we want it to. I'll take a look in webpack-bundle-analyzer, but webpack splits bundle code out, and if it's setup this way, it'll see a require in the code (the else statement), and include the map code in the main bundle, and then split off a bundle with the same code. Best case, it loads the bundle immediately, and uses it for both (which defeats the purpose of lazyloading in the first place), or it would load the bundle while also having the same code in the main build. I'll try to look and see how webpack is splitting this up

tadjohnston:
@orionstein `require.ensure` returns `require` as the parameter.  With that being said if we do:
```
require.ensure(['mapbox-gl/dist/mapbox-gl'], require => {
```
there is no way to set `this.MapboxGL` to the return by the package and subsequently fails on `accessToken` since `this.MapboxGL` is undefined.

Also, for some reason (and I am aware of the `require`), webpack doesn't seem to have an issue with the way it's done here.  But just to make sure, cause @tcwilson raised the same point, I'll just remove it.

orionstein:
Ah, right, it's
require.ensure(['mapbox-gl/dist/mapbox-gl'], require => {
  this.MapboxGL = require('mapbox-gl/dist/mapbox-gl')
  this.setupMapbox()
})

If it looks like it's bundling correctly, I guess it's not a problem, it's just not how I had thought webpack handles bundling like this

tadjohnston:
Closing, not needed.

PR #70, opened 2017-11-02
Some changes for RatingBar after first use

paulpostel:
[Feature](https://www.pivotaltracker.com/story/show/152405882)

This PR makes some changes to the RatingBar components:

- Changes Star to use passed-in className prop
- Adds new required prop `id` for RatingBar, the ids being generated for rating items were not globally unique. this broke RatingBar when more than one of them was rendered in a page.
- Adds an optional second color for RatingsBar, this enables control over both sides of a partially-filled `RatingItem` (see `Two-color Rating` in Storybook => Ratings)

### DIFF
--- a/packages/react-ui-core/src/Ratings/LinearGradient.js
+++ b/packages/react-ui-core/src/Ratings/LinearGradient.js
@@ -3,9 +3,10 @@ import PropTypes from 'prop-types'
 
 export default class LinearGradient extends Component {
   static propTypes = {
-    id: PropTypes.string,
-    color: PropTypes.string,
-    width: PropTypes.string,
+    id: PropTypes.string.isRequired,
+    fillColor: PropTypes.string.isRequired,
+    backgroundFillColor: PropTypes.string,
+    width: PropTypes.string.isRequired,
   }
 
   static defaultProps = {
@@ -15,17 +16,34 @@ export default class LinearGradient extends Component {
   render() {
     const {
       id,
-      color,
+      fillColor,
+      backgroundFillColor,
       width,
     } = this.props
+
+    let backgroundStopOpacity = 1
+    let effectiveBackgroundFillColor = backgroundFillColor
+
+    if (!effectiveBackgroundFillColor) {
+      effectiveBackgroundFillColor = fillColor
+      backgroundStopOpacity = 0
+    }
 
     return (
       <defs>
         <linearGradient id={id}>
-          <stop offset="0%" stopOpacity="1" stopColor={color} />
-          <stop offset={width} stopOpacity="1" stopColor={color} />
-          <stop offset={width} stopOpacity="0" stopColor={color} />
-          <stop offset="100%" stopOpacity="0" stopColor={color} />
+          <stop offset="0%" stopOpacity="1" stopColor={fillColor} />
+          <stop offset={width} stopOpacity="1" stopColor={fillColor} />
+          <stop
+            offset={width}
+            stopOpacity={backgroundStopOpacity}
+            stopColor={effectiveBackgroundFillColor}
+          />
+          <stop
+            offset="100%"
+            stopOpacity={backgroundStopOpacity}
+            stopColor={effectiveBackgroundFillColor}
+          />
         </linearGradient>
       </defs>
     )
--- a/packages/react-ui-core/src/Ratings/RatingBar.js
+++ b/packages/react-ui-core/src/Ratings/RatingBar.js
@@ -1,6 +1,7 @@
 import React, { PureComponent } from 'react'
 import PropTypes from 'prop-types'
 import themed from 'react-themed'
+
 import classnames from 'classnames'
 import Star from './Star'
 
@@ -10,8 +11,9 @@ import Star from './Star'
 
 export default class RatingBar extends PureComponent {
   static propTypes = {
+    uniqueId: PropTypes.string.isRequired,
     theme: PropTypes.object,
-    color: PropTypes.string,
+    fillColor: PropTypes.string,
     className: PropTypes.string,
     maxScore: PropTypes.number,
     score: PropTypes.number,
@@ -23,14 +25,15 @@ export default class RatingBar extends PureComponent {
     score: 0,
     maxScore: 5,
     RatingItem: Star,
+    fillColor: 'black',
   }
 
   get ratingItems() {
     const {
+      uniqueId,
       maxScore,
       RatingItem,
       theme,
-      color,
       ...props
     } = this.props
 
@@ -41,10 +44,10 @@ export default class RatingBar extends PureComponent {
           classnames(
             theme.RatingBar_Item,
             theme[`RatingBar_Item-${index}`],
-            color && theme[`RatingBar_Item-${color}`]
+            this.props.fillColor && theme[`RatingBar_Item-${this.props.fillColor}`]
           )
         }
-        id={`rating-item-${index}`}
+        uniqueId={`${uniqueId}-rating-item-${index}`}
         width={`${this.fillWidth(index)}%`}
         {...props}
       />
--- a/packages/react-ui-core/src/Ratings/Star.js
+++ b/packages/react-ui-core/src/Ratings/Star.js
@@ -4,8 +4,9 @@ import LinearGradient from './LinearGradient'
 
 export default class Star extends Component {
   static propTypes = {
-    id: PropTypes.string.isRequired,
-    color: PropTypes.string,
+    uniqueId: PropTypes.string.isRequired,
+    fillColor: PropTypes.string.isRequired,
+    backgroundFillColor: PropTypes.string,
     width: PropTypes.string,
     className: PropTypes.string,
   }
@@ -16,15 +17,16 @@ export default class Star extends Component {
 
   render() {
     const {
-      id,
-      color,
+      uniqueId,
+      fillColor,
+      backgroundFillColor,
       width,
       className,
       ...props
     } = this.props
 
     return (
-      <div {...props}>
+      <div className={className} {...props}>
         <svg
           xmlns="http://www.w3.org/2000/svg"
           width="50"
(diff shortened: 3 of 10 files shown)


packages/react-ui-core/src/Ratings/LinearGradient.js
```diff
@@ -3,9 +3,10 @@ import PropTypes from 'prop-types'
 
 export default class LinearGradient extends Component {
   static propTypes = {
-    id: PropTypes.string,
-    color: PropTypes.string,
-    width: PropTypes.string,
+    id: PropTypes.string.isRequired,
+    color: PropTypes.string.isRequired,
+    color2: PropTypes.string,
```
tadjohnston: What do you think about changing these names to `fillColor` and `backgroundFillColor`? `color` and `color2` doesn't tell me what the difference is here.

packages/react-ui-core/src/Ratings/LinearGradient.js
```diff
@@ -3,9 +3,10 @@ import PropTypes from 'prop-types'
 
 export default class LinearGradient extends Component {
   static propTypes = {
-    id: PropTypes.string,
-    color: PropTypes.string,
-    width: PropTypes.string,
+    id: PropTypes.string.isRequired,
```
tadjohnston: I don't mind making these required, but I almost wonder if `id` should be used in `context` instead.

packages/react-ui-core/src/Ratings/RatingBar.js
```diff
@@ -10,6 +10,7 @@ import Star from './Star'
 
 export default class RatingBar extends PureComponent {
   static propTypes = {
+    id: PropTypes.string.isRequired,
```
tadjohnston: Instead of making this required, how about we add something to generate one if an `id` doesn't exist.  Maybe something like this:
```

import { randomId } from '@rentpath/react-ui-utils'

get id() {
  return this.props.id || randomId()
}
```

Passing this down as `context` may fix the issues with `LinearGradient`.

packages/react-ui-core/src/Ratings/RatingBar.js
```diff
@@ -41,10 +43,10 @@ export default class RatingBar extends PureComponent {
           classnames(
             theme.RatingBar_Item,
             theme[`RatingBar_Item-${index}`],
-            color && theme[`RatingBar_Item-${color}`]
+            this.props.color && theme[`RatingBar_Item-${this.props.color}`]
```
tadjohnston: Why did this need to be changed?

packages/react-ui-core/src/Ratings/Star.js
```diff
@@ -5,7 +5,8 @@ import LinearGradient from './LinearGradient'
 export default class Star extends Component {
   static propTypes = {
     id: PropTypes.string.isRequired,
-    color: PropTypes.string,
+    color: PropTypes.string.isRequired,
+    color2: PropTypes.string,
```
tadjohnston: Same comment as [above](https://github.com/rentpath/react-ui/pull/70/files#r148677215)

packages/react-ui-core/src/Ratings/Star.js
```diff
@@ -18,13 +19,14 @@ export default class Star extends Component {
...
       color,
+      color2,
       width,
       className,
       ...props
     } = this.props
 
     return (
-      <div {...props}>
+      <div className={className} {...props}>
```
tadjohnston: 👍

packages/react-ui-core/src/Ratings/RatingBar.js
```diff
@@ -23,14 +24,15 @@ export default class RatingBar extends PureComponent {
     score: 0,
     maxScore: 5,
     RatingItem: Star,
+    color: 'royalblue',
```
tadjohnston: Do we really need this default?

packages/react-ui-core/src/Ratings/RatingBar.js
paulpostel: color was being pulled out of props, so it wasn't being passed down to the `RatingItem`

packages/react-ui-core/src/Ratings/RatingBar.js
paulpostel: Not really, but since i made it a required prop, tests stared throwing warnings.  This was the easy way out.  I'll update the tests instead if you think it's important.

tadjohnston: I just didn't want to force a color. Maybe if anything we should just default to `black` or or something.

packages/react-ui-core/src/Ratings/RatingBar.js
tadjohnston: Ah, ok, makes sense 👍

stories/examples/Ratings.js
```diff
@@ -92,9 +96,21 @@ export const Partial2 = (
...
   />
 )
+
+export const PartialTwoColor = (
+  <RatingBar
+    id="PartialTwoColor"
+    score={3.5}
+    label="150"
+    color="yellow"
+    color2="royalblue"
```
tadjohnston: These need to be updated.  Storybook shows the wrong thing.

packages/react-ui-core/src/Ratings/LinearGradient.js
```diff
@@ -15,17 +16,34 @@ export default class LinearGradient extends Component {
...
+      fillColor,
+      backgroundFillColor,
       width,
     } = this.props
 
+    let backgroundStopOpacity = 1
+    let effectiveBackgroundFillColor = backgroundFillColor
+
+    if (!effectiveBackgroundFillColor) {
+      effectiveBackgroundFillColor = fillColor
```
tadjohnston: Does it matter if this is set if opacity is 0 still?

paulpostel: I'm not really sure, the last two stops were setting color to fillColor before, so i wanted to render the same markup as it was doing previously if no backgroundFillColor was provided.

packages/react-ui-core/src/Ratings/RatingBar.js
```diff
@@ -23,14 +26,24 @@ export default class RatingBar extends PureComponent {
     score: 0,
     maxScore: 5,
     RatingItem: Star,
+    fillColor: 'black',
+  }
+
+  constructor(props) {
+    super(props)
+    this.effectiveId = this.props.id || randomId()
+  }
```
paulpostel: @tadjohnston, i'm having lots of trouble making this generated id work properly.  Here's what i see happening:  on SSR, i generate a new id, it gets passed down to `Star` where it is properly set on the `LinearGradient` and the reference is correct in the svg's `path`.  The, on the client side the constructor runs again, a new id gets generated and `LinearGradient` gets updated with it.  However, the svg's `path` does not re-render with the new id, it still has the old id.  That breaks things.   Also, isn't this causing a mismatch between client/server rendering?   If i pass in an id rather than generating one, everything works properly.

packages/react-ui-core/src/Ratings/RatingBar.js
```diff
@@ -23,14 +26,24 @@ export default class RatingBar extends PureComponent {
...
+    return this.effectiveId
   }
 
   get ratingItems() {
     const {
+      id,
       maxScore,
       RatingItem,
       theme,
-      color,
```
ninomichaels: `fillColor` needs to be added here and called accordingly below (instead of `this.props.fillColor`), right now `fillColor` is being added as an invalid attribute on `RatingItem`.

packages/react-ui-core/src/Ratings/LinearGradient.js
```diff
@@ -3,9 +3,10 @@ import PropTypes from 'prop-types'
 
 export default class LinearGradient extends Component {
   static propTypes = {
-    id: PropTypes.string,
-    color: PropTypes.string,
-    width: PropTypes.string,
+    id: PropTypes.string.isRequired,
+    fillColor: PropTypes.string.isRequired,
```
ninomichaels: getting some warnings about these in other places `LinearGradient` is being used

packages/react-ui-core/src/Ratings/RatingBar.js
```diff
@@ -41,10 +54,10 @@ export default class RatingBar extends PureComponent {
...
             theme.RatingBar_Item,
             theme[`RatingBar_Item-${index}`],
-            color && theme[`RatingBar_Item-${color}`]
+            this.props.fillColor && theme[`RatingBar_Item-${this.props.fillColor}`]
           )
         }
-        id={`rating-item-${index}`}
+        id={`${this.id}-rating-item-${index}`}
         width={`${this.fillWidth(index)}%`}
         {...props}
```
ninomichaels: `fillColor` is not being passed down into some of the other examples so is causing it to be added at this level as an invalid attribute.

stories/examples/Ratings.js
```diff
@@ -74,27 +76,41 @@ export const ManyRatings = (
...
+    id="Partial3"
     score={0.75}
     label="150"
-    color="royalblue"
+    fillColor="royalblue"
+    theme={RatingBarTheme}
+  />
+)
+
+export const PartialTwofillColor = (
```
ninomichaels: looks like this needs to changed to `PartialTwoColor`

paulpostel:
After further consideration, i don't see how a generated random id for the link between the svg and its gradient can work.  It will necessarily generate a different id on the server and client, and thus different markup.   I tried to make the id a function of the `Star`'s props and that works client/server, but it generates duplicate ids across/within ratings bars.  In the end I decided requiring a unique id prop in `RatingBar` was the best solution.  I renamed it `uniqueId` because if i were to pass an `id` prop to a component, my expectation would be for that id to appear on an element.  That doesn't happen here.
cc @lfender6445 

@ninomichaels, i believe I have all the examples cleaned up now.

paulpostel:
FYI, you will see a warning issued on SSR about a `matching linearGradient`, this is a known react bug:  https://github.com/facebook/react/issues/11354

lfender6445:
warning will be fixed in next release of react it looks like

👍

ninomichaels:
LGTM 👍

PR #71, opened 2017-11-18
adding brand repos

tadjohnston:
No Story

This is setup for React brand specific components for AG and Rent.

Also, `react-ui-map` tests weren't passing. This was due to `require.ensure` not being available in the tests.  After some research I found out why we can't use a dynamic import instead.  Basically it's due to the fact that babel is transpiling it to a straight `require` which is causing webpack to not split on it.  

Solution: Added `dynamic-import-webpack` for the production build which transforms the dynamic import into a `require.ensure`.  For testing we're now using the dynamic import.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,9 @@
       "packages/react-ui-core",
       "packages/react-ui-map",
       "packages/react-ui-tracking",
-      "packages/react-ui-utils"
+      "packages/react-ui-utils",
+      "packages/react-ui-ag",
+      "packages/react-ui-rent"
     ]
   },
   "devDependencies": {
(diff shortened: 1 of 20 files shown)


lfender6445:
seeing this in storybook

screen shot 2017-11-27 at 1 36 44 pm

i've run yarn install for the map as well and rebuilt

tadjohnston:
Fixed.  It was due to not including `babel-plugin-dynamic-import-webpack` in the base package along with the storybook `.babelrc`.

lfender6445:
👍

pkanuganti:
LGTM 👍

PR #72, opened 2017-11-26
Mapbox upgrade

lfender6445:
Fixes https://github.com/mapbox/mapbox-gl-js/issues/4582 for me 

and https://github.com/mapbox/mapbox-gl-js/issues/4846

### DIFF
--- a/packages/react-ui-map/package.json
+++ b/packages/react-ui-map/package.json
@@ -27,7 +27,7 @@
   },
   "dependencies": {
     "classnames": "^2.2.5",
-    "mapbox-gl": "0.40.1",
+    "mapbox-gl": "^0.42.2",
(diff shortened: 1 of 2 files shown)


lfender6445:
cc @tadjohnston @kevinolson @paulpostel just letting you know im merging this asap because of the pressure to get search as i move out

this helps with an issue calling map.fitbounds() with padding (which sometimes causes a tinier version of the map to appear)

lfender6445:
turns out well need more time to upgrade. looks like there were some new artifacts introduced with icon-overlap that i need to spend more time researching so i'm keeping the rjs map where its at.

tadjohnston:
You're not supposed to bump these versions manually. Please use lerna to do that for you in the future.

PR #75, opened 2017-11-30
Create List, List Item and List Card components

bethstevic:
[Card](https://rentpath.leankit.com/card/567623822)


packages/react-ui-core/src/List/List.js
```diff
@@ -0,0 +1,49 @@
...
+      theme,
+      className,
+      Item,
+      items,
+      listItemNodeType,
+      ...props
+    } = this.props
+
+    return (
+      <this.props.nodeType
```
tadjohnston: Let's destructure this above.  Reason being is for 2 things:
1. We don't have to look it up twice
2.  It then doesn't get passed down to `<Item />`

packages/react-ui-core/src/List/List.js
```diff
@@ -0,0 +1,49 @@
...
+        className={cn(
+          theme.List,
+          className,
+        )}
+      >
+        {items.map((item, i) => (
+          <Item key={i} listItemNodeType={listItemNodeType} {...props}>
```
tadjohnston: Can we see if we can make the `key` more unique then this?  Making it the index can potentially cause issues if we removed an item / spliced from this list.  Typically it's not a huge issue, but since this component is generic, we want to make sure if someone does remove an item, it doesn't have a weird side effect with it.

packages/react-ui-core/src/List/ListItem.js
```diff
@@ -0,0 +1,45 @@
...
+
+@themed('*', { pure: true })
+
+export default class ListItem extends PureComponent {
+
+  static propTypes = {
+    theme: PropTypes.object,
+    className: PropTypes.string,
+    orientation: PropTypes.string,
+    listItemNodeType: PropTypes.string,
```
tadjohnston: Let's rename this to just `nodeType`. Fixing what I suggest [above](https://github.com/rentpath/react-ui/pull/75/files#diff-955b1a65301832dda07a85d377a06f87R35) should rectify this prop name.

packages/react-ui-core/src/List/ListItem.js
```diff
@@ -0,0 +1,45 @@
...
+    const {
+      orientation,
+      theme,
+      className,
+      children,
+      ...props
+    } = this.props
+
+    return (
+      <this.props.listItemNodeType
```
tadjohnston: Let's destructure this above so it doesn't get passed into the spread props below.

stories/theme/ListCard.css
```diff
@@ -0,0 +1,6 @@
+.container {
```
tadjohnston: I don't think you meant to keep this file right?

stories/index.js
```diff
@@ -91,3 +95,9 @@ storiesOf('Ratings', module)
   .addWithInfo('Partial Rating 2', 'Partial Rating 2', () => Partial2)
   .addWithInfo('Partial Rating 3', 'Partial Rating 3', () => Partial3)
   .addWithInfo('Two-color Rating', 'Two-color Rating', () => PartialTwoColor)
+
+storiesOf('List', module)
+  .addWithInfo('default', 'Default List', () => DefaultList)
+  .addWithInfo('List with passed node types', 'Passed Node Types', () => ListWithPassedNodeTypes)
+  .addWithInfo('List with own item component', 'Own item component', () => ListWithOwnItemComponent)
+  .addWithInfo('List with Listing Card', 'Listing Card', () => ListWithListingCard)
```
tadjohnston: This should probably be an orientation example yes?

packages/react-ui-core/src/List/__tests__/List-test.js
```diff
@@ -0,0 +1,43 @@
...
+
+  it('renders a List', () => {
+    wrapper.setProps({ theme })
+    expect(wrapper.find(List).length).toBe(1)
+  })
+
+  it('applies a theme classname', () => {
+    wrapper.setProps({ theme })
+    const className = wrapper.find('ul').prop('className')
+    expect(className).toContain('List')
```
tadjohnston: `toContain` here will match `List`, `List-orientation` and anything else with List in it, even `Foo-List-Foo`.  Are you sure you want to test this?

packages/react-ui-core/src/List/__tests__/List-test.js
```diff
@@ -0,0 +1,43 @@
...
+
+  it('accepts custom node type', () => {
+    wrapper.setProps({ nodeType: 'div', theme })
+    expect(wrapper.find('div').length).toBe(1)
+  })
+
+  it('accepts custom orientation', () => {
+    wrapper.setProps({ orientation: 'horizontal', theme })
+    const className = wrapper.find('ul').prop('className')
+    expect(className).toContain('horizontal')
```
tadjohnston: You probably want `toContain('List-horizontal')` just so we know that exact class exists.

packages/react-ui-core/src/List/List.js
```diff
@@ -0,0 +1,81 @@
...
+@themed('*', { pure: true })
+
+export default class List extends PureComponent {
+  static propTypes = {
+    theme: PropTypes.object,
+    className: PropTypes.string,
+    nodeType: PropTypes.string,
+    Item: PropTypes.oneOfType([
+      PropTypes.node,
+      PropTypes.string,
```
nray621: Nitpick, and one I'm not too sure about, but it seems that, according to the docs, `PropTypes.node` includes strings. 
https://reactjs.org/docs/typechecking-with-proptypes.html

nray621:
👍 LGTM

PR #76, opened 2017-11-30
Add data-tids for counter

jtheodore-rent:
This PR adds support for a custom data-tid to be placed where the text is displayed for the value of the counter element.

Needed for https://github.com/rentpath/rent-js/pull/369

Author: JT
[Card](https://rentpath.leankit.com/card/583456045)

### DIFF
--- a/packages/react-ui-core/src/Counter/Counter.js
+++ b/packages/react-ui-core/src/Counter/Counter.js
@@ -86,14 +86,19 @@ export default class Counter extends PureComponent {
       theme,
       label,
       className,
+      decrementOperator,
+      incrementOperator,
+      ...props
     } = this.props
 
     return (
-      <div className={
-        classnames(
-          theme.Counter,
-          className
-        )}
+      <div
+        className={
+          classnames(
+            theme.Counter,
+            className
+          )}
+        {...props}
       >
         {label && this.renderLabel}
         <div className={theme.Counter_Controls}>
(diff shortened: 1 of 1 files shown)


tadjohnston:
You don't need to do this. I can show you a better solution.

packages/react-ui-core/src/Counter/Counter.js
```diff
@@ -20,6 +20,7 @@ export default class Counter extends PureComponent {
     min: PropTypes.number,
     max: PropTypes.number,
     className: PropTypes.string,
+    textTid: PropTypes.string,
```
lfender6445: lets call this tid as a generic prop for test id

jtheodore-rent: The reason I called it textTid is that it specifically refers to the text that will display for this instance of the counter component - not a generic tid that the whole component would use.

I guess I could give it a generic tid name, though, and then derive the text data-tid from that

tadjohnston: We shouldn't be adding any specific nightwatch related data to these components. That's not what react-ui-core is about.  The better thing to do here would be to _not_ pass in a `data-tid`, but rather spread the props on the outer most div.  That way the component (without having to explicitly pass it) now has the `data-tid`.  If you _really_ need something on a child node, then we can add something more static like `data-counter="text"` or something. I can talk this over when you're ready.

lfender6445: oh i didn't realize that - ideally the TID sits on the outside and you use that to climb into the HTML so we are not adding attributes needlessly - is this possible? 

eg `document.querySelector(someTid).text()`

lfender6445:
👍

PR #78, opened 2017-12-05
Nr/ya/searchbar

yasabere:
Wip


packages/react-ui-core/src/List/List.js
```diff
@@ -70,7 +70,13 @@ export default class List extends PureComponent {
         )}
       >
         {items.map((item, i) => (
-          <Item key={this.itemId(i)} nodeType={listItemNodeType} {...props}>
+          <Item
+            key={this.itemId(i)}
+            index={i}
+            nodeType={listItemNodeType}
+            theme={theme}
```
tadjohnston: Why are you passing theme here?

packages/react-ui-core/src/List/List.js
```diff
@@ -70,7 +70,13 @@ export default class List extends PureComponent {
         )}
       >
         {items.map((item, i) => (
-          <Item key={this.itemId(i)} nodeType={listItemNodeType} {...props}>
+          <Item
+            key={this.itemId(i)}
+            index={i}
```
tadjohnston: What is this for?

stories/examples/Searchbar.js
```diff
@@ -0,0 +1,94 @@
...
+    )
+  }
+}
+
+export const DefaultSearchBar = (
+  <Example />
+)
+
+export const SearchBarWithHTMLButton = (
+  [<SearchBar
```
tadjohnston: You don't need this brackets around the component.

stories/examples/Searchbar.js
```diff
@@ -0,0 +1,94 @@
...
+        return value.props.children
+      }
+      return value
+    }}
+  />,
+  ]
+)
+
+export const SearchBarSubmitButton = (
+  [<SearchBar
```
tadjohnston: You don't need this brackets around the component.

packages/react-ui-core/src/SearchBar/SearchBar.js
```diff
@@ -0,0 +1,214 @@
...
+    suggestionsVisible: false,
+  }
+
+  constructor(props) {
+    super(props)
+
+    const { value, suggestionsVisible } = this.props
+
+    this.reset = this.reset.bind(this)
+    this.handleChange = this.handleChange.bind(this)
```
tadjohnston: Seems like there is way too much going on with binding. I wonder if we really need all of these?

nray621: I will double check, but I fairly sure those were added one by one as needed. These are mostly passed down to child components through props and need proper context when invoked.

stories/examples/Searchbar.js
nray621: Yeah... not sure why we had those 👍

packages/react-ui-core/src/List/List.js
nray621: Not sure. Will fix.

nray621: It isn't present in the spread `props` because it was destructed, and we needed access to it in 'SearchBar/ItemWrapper'

```
render() {
    const {
      selectedIndex,
      index,
      theme,
      children,
      onSelectionClick,
    } = this.props

    return (
      <div
        className={cn(
          theme.ListItem,
          selectedIndex === index && theme.Selected
        )}
        onClick={onSelectionClick}
        onMouseEnter={this.handleHover}
      >
        {children}
      </div>
    )
  }
```

mneumark:
Closing. Please delete the branch when you are done referencing it.

PR #79, opened 2017-12-05
adding data-tid to close button for modal

tadjohnston:
[Story](https://rentpath.leankit.com/card/583330299)

- added data-tid for close button
- moved spread of props so data-tid can be passed to outermost level


jtheodore-rent:
👍

PR #82, opened 2017-12-05
Bs/583475844/card component

bethstevic:
Create generic card component


packages/react-ui-core/src/Card/Card.js
```diff
@@ -0,0 +1,55 @@
...
+        className={cn(
+          theme.Card,
+          className,
+        )}
+        {...props}
+      >
+        <TitleNodeType className={theme.title}>
+          {title}
+        </TitleNodeType>
+        <ChildrenNodeType className={theme.content}>
```
tadjohnston: `theme` classes should be uppercased.

packages/react-ui-core/src/Card/Card.js
```diff
@@ -0,0 +1,55 @@
...
+    title: PropTypes.string,
+    nodeType: PropTypes.string,
+    titleNodeType: PropTypes.string,
+    contentNodeType: PropTypes.string,
+    content: PropTypes.node,
+  }
+
+  static defaultProps = {
+    theme: {},
+    nodeType: 'div',
```
tadjohnston: I think `nodeType` and `contentNodeType` might be a little excessive for now.

packages/react-ui-core/src/Card/Card.js
```diff
@@ -0,0 +1,55 @@
...
+
+    return (
+      <NodeType
+        className={cn(
+          theme.Card,
+          className,
+        )}
+        {...props}
+      >
+        <TitleNodeType className={theme.title}>
```
tadjohnston: I think this should probably be its own component maybe?

packages/react-ui-core/src/Card/Card.js
```diff
@@ -0,0 +1,55 @@
...
+          {title}
+        </TitleNodeType>
+        <ChildrenNodeType className={theme.content}>
+          {content}
```
tadjohnston: It would probably be better to use `children` here.

stories/examples/Card.js
```diff
@@ -0,0 +1,25 @@
...
+    title={title}
+    content={cardContent}
+  />
+)
+
+export const CardWithPassedNodeTypes = (
+  <Card
+    theme={CardTheme}
+    title={title}
+    content={cardContent}
```
tadjohnston: Content is really the children correct?

tadjohnston:
I'm not sure if creating a generic card here will do us any good when creating listing cards.  While, what you have is a good starting base, it's not providing enough for even a generic component IMO.  It seems like there's just way too much flexibility for every node being a customizable type and no real structure that dictates a card. Perhaps starting a listing card in react-ui-rent would be better and then see if there's anything we can break down from there into generic components.  For example, title would probably be a good example of something that could be generic and its own component.

```
// Title
static defaultProps = {
  nodeType: 'h2'
}

render() {
  return(
    <NodeType 
      className={classnames(
        theme.Title,
        className
      )}
      data-tid={title}
    >
      {children}
    </NodeType>
  )
}
```

Creating a generic card is a pretty tough thing and I know I said start on the lowest component possible and work your way back out, but we may need to approach this card differently to give us a better idea.

grenti:
LGTM 👍

packages/react-ui-core/src/Card/__tests__/Card-test.js
```diff
@@ -0,0 +1,29 @@
...
+
+  it('applies a theme classname', () => {
+    wrapper.setProps({ theme })
+    const className = wrapper.prop('className')
+    expect(className).toEqual('Card')
+  })
+
+  it('renders passed children and adds data-tid', () => {
+    wrapper.setProps({ children: content })
+    expect(wrapper.find('div').at(1).html()).toBe('<div data-tid="card-content"><h1>Hello world</h1></div>')
```
tadjohnston: You can use a snapshot here instead.
```
import renderer from 'react-test-renderer'

it('renders passed children and adds data-tid', () => {
    const snap = renderer
      .create(<Card>{content}</Card>)
      .toJSON()
    expect(snap).toMatchSnapshot()
})
```

packages/react-ui-core/src/Card/__tests__/Card-test.js
```diff
@@ -0,0 +1,29 @@
...
+const Card = ThemedCard.WrappedComponent
+const content = <h1>Hello world</h1>
+
+describe('Card', () => {
+  let wrapper
+
+  beforeEach(() => {
+    wrapper = shallow(
+      <Card />,
+      { lifecycleExperimental: true }
```
tadjohnston: @mneumark found based on our enzyme version we actually don't need this. So we can remove it 😄

packages/react-ui-core/src/Card/Card.js
```diff
@@ -0,0 +1,42 @@
...
+export default class Card extends PureComponent {
+  static propTypes = {
+    theme: PropTypes.object,
+    className: PropTypes.string,
+    children: PropTypes.node,
+  }
+
+  static defaultProps = {
+    theme: {},
+    titleNodeType: 'h2',
```
tadjohnston: If we're not using this you can remove it.

stories/examples/Card.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react'
+import { Card, List } from 'react-ui-core/src'
+import { CardTheme } from '../theme'
+
+const cardContent = <List items={['One', 'Two', 'Three']} />
+
+export const DefaultCard = (
+  <Card
```
tadjohnston: I'd like to see us getting in the habbit of using the `Theme` / `ThemeProvider` from `react-themed`.
```
import { Theme } from 'react-themed'
import { CardTheme } from './theme'

const theme = compose({}, CardTheme)

export const DefaultCard = (
  <Theme theme={theme}>
    <Card/>
  </Theme>
)
```

It then assures we properly have the decorator classes applied correctly.

PR #85, opened 2017-12-07
Add RadioGroup and RadioButton Components

mneumark:
[Story](https://rentpath.leankit.com/card/583855514)

Paired with Anye

* created RadioButton component
* created RadioGroup component
* updated to React 16.2.0


packages/react-ui-core/src/Form/RadioButton.js
```diff
@@ -0,0 +1,47 @@
...
+import Field from './Field'
+
+@themed('*', {
+  pure: true,
+})
+export default class RadioButton extends PureComponent {
+    static propTypes = {
+      className: PropTypes.string,
+      orientation: PropTypes.string,
+      hideInput: PropTypes.bool,
```
bethstevic: Just a nitpick, but I really don't like the name hideInput. if I was just trying to use this without knowing anything about it I don't think I would understand that hideInput is for the radio circle

bethstevic:
Other than the naming nitpick, looks good to me 👍

packages/react-ui-core/src/Form/RadioButton.js
grenti: So I think the logic behind hideInput is that we are hiding the input element (of type radio). Also we wanted to stay close to html element nomenclature as possible, which is why we have checked (based on html input checked) and not selected. @mneumark and @tadjohnston what do you think?

tadjohnston: I'm good with whatever decision you guys make 😄

mneumark: I'm pretty neutral. Let me ask this. Do you think we will end up in a situation a lot where the radio buttons need to change from hidden to unhidden? Or is it all a one and done sort of thing?

PR #86, opened 2017-12-07
Rent mobile lead form changes

tcwilson:
This also fixes a currently failing test on master

[Card](https://rentpath.leankit.com/card/577968668)


packages/react-ui-core/src/Form/__tests__/mocks/theme.js
```diff
@@ -23,5 +23,6 @@ export default keyMirror([
   'RadioButton-checked',
   'RadioButton-hide',
   'RadioButton-vertical',
+  'RadioButton-hidden',
```
tadjohnston: There's already a `RadioButton-hide`. Can we remove that one?

packages/react-ui-rent/src/LeadForm/Fields/Message.js
```diff
@@ -25,39 +25,23 @@ export default class Message extends PureComponent {
...
       label,
       ...props
     } = this.props
 
     return (
-      <Field
-        label={this.label}
-        {...props}
-      />
+      <div className={theme.Message}>
```
tadjohnston: Why did you need to remove label here?

packages/react-ui-rent/src/LeadForm/Fields/OptInNewsLetter.js
```diff
@@ -1,27 +1,57 @@
...
       PropTypes.object,
       PropTypes.func,
     ]),
+    defaultChecked: PropTypes.bool,
   }
 
   static defaultProps = {
+    theme: {},
     name: 'opt_in_newsletter',
+    id: 'lead_form_opt_in_newsletter',
```
tadjohnston: If I have multiple lead forms on the page with an `OptInNewsLetter`, how does distinguish between the two? I guess I'm not understanding what the purpose of this `id` is here.

packages/react-ui-rent/src/LeadForm/Fields/OptInNewsLetter.js
```diff
@@ -1,27 +1,57 @@
...
+
     return (
-      <Field {...this.props} />
+      <div className={theme.OptInNewsLetter}>
+        <Field
+          id={id}
+          {...props}
+        />
+        {/* eslint-disable jsx-a11y/label-has-for */}
+        <label
```
tadjohnston: I'm not getting why you moved Label out.  Also, there is a `Label` component you should be using instead.

packages/react-ui-rent/src/LeadForm/Fields/OptInNewsLetter.js
tadjohnston: If you're trying to use this for `htmlFor`, then please take a look at https://github.com/rentpath/react-ui/blob/master/packages/react-ui-utils/src/randomId.js.  You can look in react-ui-core because it has been used there, especially in regards to `Field`.

packages/react-ui-rent/src/LeadForm/Fields/OptInNewsLetter.js
```diff
@@ -1,27 +1,57 @@
...
+    theme: PropTypes.object,
     name: PropTypes.string,
+    id: PropTypes.string,
     type: PropTypes.string,
     label: PropTypes.oneOfType([
       PropTypes.string,
       PropTypes.object,
       PropTypes.func,
     ]),
+    defaultChecked: PropTypes.bool,
```
tadjohnston: Did you want this to be a controlled or uncontrolled component?

lfender6445:
Are the requirements for Message field differently? I would expect `Message` label to disappear as well

screen shot 2017-12-07 at 3 10 12 pm

fredleveau:
yeah, it would technically disappear as well but users probably don't change the canned message (data?) and therefore would not disappear since there is no new input in the field.

how about:
WHEN user does not edit the canned message
THEN nothing changes and message label remains

WHEN user edits canned message
THEN message label dissapears
AND field behaves in similar manner to other fields

Thoughts?

tcwilson:
@tadjohnston want to give it another look?

packages/react-ui-rent/src/LeadModal/LeadModal.js
```diff
@@ -27,7 +42,12 @@ export default class LeadModal extends Component {
       <Modal
         isOpen={isOpen}
         CloseButton={CloseButton}
+        onClose={onClose}
```
tadjohnston: 👍

packages/react-ui-rent/src/LeadModal/CloseButton.js
```diff
@@ -34,6 +36,9 @@ export default class CloseButton extends PureComponent {
           theme.CloseButton
         )}
         {...props}
+        onClick={onClick}
```
tadjohnston: You don't need to pass this. The whole point of react-ui-core Modal is so you don't have to do this. See: https://github.com/rentpath/react-ui/blob/master/packages/react-ui-core/src/Modal/ModalBody.js#L41. Passing props through here accomplishes what you need.

packages/react-ui-rent/src/LeadForm/LeadForm.js
```diff
@@ -74,17 +75,23 @@ export default class LeadForm extends Component {
...
+      fields,
+      messagePlaceholder,
+    } = this.props
 
     return fields.map(fieldComposition => {
       const { field, ...props } = fieldComposition
       const name = props.name
       const FormField = field || FIELD_MAPPING[name] || Field
+      const isMessage = name === 'message'
+      const placeholder = isMessage ? { placeholder: messagePlaceholder } : {}
```
tadjohnston: I'd rather put this logic in the Message component if it makes sense.

packages/react-ui-rent/src/LeadForm/Fields/RequiredField.js
```diff
@@ -0,0 +1,66 @@
...
+    const {
+      theme,
+      placeholder,
+      ...props
+    } = this.props
+
+    return (
+      <Field
+        className={theme.RequiredField}
+        theme={theme}
```
tadjohnston: You don't need to pass this down.

packages/react-ui-rent/src/LeadForm/Fields/MoveInDate.js
```diff
@@ -0,0 +1,24 @@
...
+    name: PropTypes.string,
+    type: PropTypes.string,
+    placeholder: PropTypes.string,
+  }
+
+  static defaultProps = {
+    name: 'date',
+    type: 'date',
+    placeholder: 'Move in date',
+    theme: {},
```
tadjohnston: This prop isn't needed here.

lfender6445:
👍

packages/react-ui-rent/src/LeadForm/Fields/RequiredField.js
tcwilson: I do in this case. `props` vs `this.props`. The `placeholder` is filtered out above and theme is used in `className`

tcwilson:
@tadjohnston want to give it another shot?

packages/react-ui-rent/src/LeadForm/Fields/RequiredField.js
tadjohnston: That's fine you still don't need it.  Field is already themed.

PR #87, opened 2017-12-07
Create generic DropDown component

nray621:
[Card](https://rentpath.leankit.com/card/586156982)

Create and write tests for DropDown

### DIFF
--- a/packages/react-ui-core/src/DropDown/DefaultAnchor.js
+++ b/packages/react-ui-core/src/DropDown/DefaultAnchor.js
@@ -1,0 +1,48 @@
+import React, { PureComponent } from 'react'
+import PropTypes from 'prop-types'
+import { Button } from '../Button'
+
+export default class DefaultAnchor extends PureComponent {
+  static propTypes = {
+    toggleVisibilty: PropTypes.func,
+    handleDocumentClick: PropTypes.func,
+    text: PropTypes.node,
+    className: PropTypes.string,
+    onClick: PropTypes.func,
+  }
+
+  constructor(props) {
+    super(props)
+    this.handleClick = this.handleClick.bind(this)
+  }
+
+  componentDidMount() {
+    document.addEventListener('click', this.props.handleDocumentClick)
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('click', this.props.handleDocumentClick)
+  }
+
+  handleClick() {
+    this.props.toggleVisibilty()
+  }
+
+  render() {
+    const {
+      className,
+      text,
+      onClick, // eslint-disable-line no-unused-vars
+    } = this.props
+
+    return (
+      <Button
+        data-tid="dropdown-anchor-button"
+        onClick={this.handleClick}
+        className={className}
+      >
+        {text}
+      </Button>
+    )
+  }
+}
--- a/packages/react-ui-core/src/DropDown/DropDown.js
+++ b/packages/react-ui-core/src/DropDown/DropDown.js
@@ -1,0 +1,92 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+import themed from 'react-themed'
+import classnames from 'classnames'
+import DefaultAnchor from './DefaultAnchor'
+
+@themed(/^DropDown/)
+
+export default class DropDown extends Component {
+  static propTypes = {
+    className: PropTypes.string,
+    visible: PropTypes.bool,
+    theme: PropTypes.object,
+    children: PropTypes.node,
+    Anchor: PropTypes.oneOfType([
+      PropTypes.node,
+      PropTypes.func,
+    ]),
+  }
+
+  static defaultProps = {
+    visible: false,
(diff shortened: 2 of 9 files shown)


bethstevic:
Looks good to me 👍

grenti:
LGTM 👍

packages/react-ui-core/src/DropDown/DropDown.js
```diff
@@ -0,0 +1,69 @@
...
+    const dropDownVisible = this.state.visible
+
+    return (
+      <div
+        className={classnames(
+          theme.DropDown,
+          className
+        )}
+      >
+        <Anchor
```
tadjohnston: How does `Anchor` work if I don't pass in the prop?

stories/examples/example-helpers/DropDownAnchorButton.js
```diff
@@ -0,0 +1,16 @@
...
+  static propTypes = {
+    toggleVisibilty: PropTypes.func,
+  }
+
+  handleClick = () => {
+    this.props.toggleVisibilty()
+  }
+
+  render() {
+    return <button onClick={this.handleClick}>click me</button>
```
tadjohnston: Please use the react-ui-core `Button` instead.

packages/react-ui-core/src/DropDown/DropDown.js
nray621: I've add a default anchor per our discussion

packages/react-ui-core/src/DropDown/DropDown.js
```diff
@@ -0,0 +1,96 @@
...
+      Anchor,
+      className,
+      ...props
+    } = this.props
+
+    const dropDownVisible = this.state.visible
+
+    return (
+      <div
+        ref={ref => { this.ref = ref }}
```
tadjohnston: Can you please call this something other than `ref`? Maybe `this.dropdown`?

packages/react-ui-core/src/DropDown/DefaultAnchor.js
```diff
@@ -0,0 +1,40 @@
...
+    text: PropTypes.node,
+    className: PropTypes.string,
+  }
+
+  constructor(props) {
+    super(props)
+    this.handleClick = this.handleClick.bind(this)
+  }
+
+  componentDidMount() {
```
tadjohnston: Do we need this addEventListener here and on its parent?

packages/react-ui-core/src/DropDown/DropDown.js
```diff
@@ -0,0 +1,96 @@
...
+
+  constructor(props) {
+    super(props)
+    this.state = { visible: this.props.visible }
+    this.handleDocumentClick = this.handleDocumentClick.bind(this)
+    this.toggleVisibilty = this.toggleVisibilty.bind(this)
+  }
+
+  componentDidMount() {
+    document.addEventListener('click', this.hanldeDocumentClick)
```
tadjohnston: Spelling mistake on `this. hanldeDocumentClick`.

packages/react-ui-core/src/DropDown/DropDown.js
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('click', this.hanldeDocumentClick)
```
tadjohnston: Spelling mistake on `this. hanldeDocumentClick`.

packages/react-ui-core/src/DropDown/DropDown.js
```diff
@@ -0,0 +1,96 @@
...
+        )}
+      >
+        <Anchor
+          {...props}
+          handleDocumentClick={this.handleDocumentClick}
+          toggleVisibilty={this.toggleVisibilty}
+          dropDownVisible={dropDownVisible}
+          className={classnames(
+            theme.DropDown_Anchor,
+            theme[`DropDown_Anchor_${this.visibleTheme}`],
```
tadjohnston: `visibleTheme` is probably a bad name. It's not a "theme" per se, it's just a modifier for a class.  Can't you just do something like this without needing the extra function:

```
theme[`DropDown_Anchor-${dropDownVisible ? 'visible' : 'hidden'}`]
```
Also, since it is a modifier, you should be using `-` instead of `_`.

packages/react-ui-core/src/DropDown/DefaultAnchor.js
nray621: Removed from parent. Didn't mean to leave it there.

packages/react-ui-core/src/DropDown/DropDown.js
nray621: 👍

packages/react-ui-core/src/DropDown/DropDown.js
nray621: 👍

PR #88, opened 2017-12-07
Title Component for react-ui-core

grenti:
[Card](https://rentpath.leankit.com/card/585464788)

* Created Title component


stories/index.js
```diff
@@ -44,7 +44,9 @@ import {
   ListWithOwnItemComponent,
   DefaultLeadModal,
   DefaultCard,
+  DefaultTitle,
 } from './examples'
+import { EmbeddedLinkWithinTitle } from './examples/Title'
```
bethstevic: Can't you import this directly from examples like the other ones? Since you are exporting it from the examples index you will have access to it

stories/examples/Title.js
```diff
@@ -0,0 +1,29 @@
...
+
+const theme = compose(
+  {},
+  TitleTheme,
+)
+
+export const DefaultTitle = (
+  <ThemeProvider theme={theme}>
+    <Title
+      nodeType="h2"
```
mneumark: If this is `DefaultTitle` you probably shouldn't pass `nodeType` and leave that for the other story.

grenti: Good point - I made the changes and updated the children propType to accommodate it.

mneumark: You should still not pass the `nodeType` in this example as it looks like you are not using the default behavior in the default example.

PR #93, opened 2017-12-09
Sentry error - duplicate click

lfender6445:
# bug

https://sentry.io/rentpath/rent-js/issues/414900919/

https://rentpath.leankit.com/card/586265862

The onClick for div was passing the event where wed prefer to have complete control over the onClick handler for the Counter by limiting it to the incrementer and decrementer buttons.

### DIFF
--- a/packages/react-ui-core/package.json
+++ b/packages/react-ui-core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@rentpath/react-ui-core",
-  "version": "1.4.0",
+  "version": "1.4.1",
   "main": "lib/index.js",
   "license": "MIT",
   "scripts": {
--- a/packages/react-ui-core/src/Counter/Counter.js
+++ b/packages/react-ui-core/src/Counter/Counter.js
@@ -109,6 +109,8 @@ export default class Counter extends PureComponent {
             className
           )}
         {...props}
+        // NOTE: ensure click is noop so handler is only on counter spans
+        onClick={() => {}}
       >
         {label && this.renderLabel}
         <div className={theme.Counter_Controls}>
--- a/packages/react-ui-core/src/Counter/__tests__/Counter-test.js
+++ b/packages/react-ui-core/src/Counter/__tests__/Counter-test.js
@@ -80,6 +80,7 @@ describe('Counter', () => {
 
   describe('onClick', () => {
     it('maintains count between min and max', () => {
+      const onClick = jest.fn()
       const wrapper = mount(
         <Counter
           min={1}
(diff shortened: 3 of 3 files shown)


packages/react-ui-core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@rentpath/react-ui-core",
-  "version": "1.4.0",
+  "version": "1.4.1",
```
tadjohnston: Please do _not_ bump this version here. lerna should be handling this.

packages/react-ui-core/src/Counter/Counter.js
```diff
@@ -109,6 +109,8 @@ export default class Counter extends PureComponent {
             className
           )}
         {...props}
+        // NOTE: ensure click is noop so handler is only on counter spans
+        onClick={() => {}}
```
tadjohnston: If this is what we really want to do, I think it would be more appropriate to add it to the destructure above instead.  Doing this here creates an unnecessary `onClick` event on every counter.

packages/react-ui-core/src/Counter/__tests__/Counter-test.js
```diff
@@ -101,6 +102,23 @@ describe('Counter', () => {
...
+        <Counter
+          min={1}
+          count={1}
+          max={3}
+          theme={theme}
+          className="wrappingDiv"
+          onClick={onClick}
+        />
+      )
+      const outerDiv = wrapper.find('div.wrappingDiv')
```
tadjohnston: I don't think you need to `find` anything.  By default the wrapper should be the root node which is what you want to simulate.  I think you can just do the following instead:
```
  wrapper.simulate('click')
  expect(wrapper.prop('onClick')).not.toBeCalled()
})
```

packages/react-ui-core/src/Counter/__tests__/Counter-test.js
```diff
@@ -101,6 +102,23 @@ describe('Counter', () => {
       expect(wrapper.state('count')).toEqual(3)
     })
 
+    it('does not allow clicks to propogate from outer div', () => {
+      const onClick = jest.fn()
+      const wrapper = mount(
```
tadjohnston: Any reason not to use `shallow` here?

tadjohnston:
I'm not sure I follow on this bug.  If I'm understanding correctly, something external to this component is passing an `onClick` prop to the Counter?  If that's the case, why aren't we fixing this in the component that's passing `onClick` in?  If we really need to fix it here, please see my (comment)[https://github.com/rentpath/react-ui/pull/93#discussion_r155984617]

lfender6445:
closing in preference of https://github.com/rentpath/react-ui/pull/98

PR #95, opened 2017-12-11
Create Filter Card

mneumark:
[Story](https://rentpath.leankit.com/card/587257279)

* Create FilterCard component
* create related examples
* set up some structure for react-ui-ag
* add react-ui-core to react-ui-ag
* added data-tid to core Text, Button and Title

### DIFF
--- a/packages/react-ui-ag/src/Filters/FilterCard.js
+++ b/packages/react-ui-ag/src/Filters/FilterCard.js
@@ -13,16 +13,14 @@ export default class FilterCard extends PureComponent {
     className: PropTypes.string,
     theme: PropTypes.object,
     children: PropTypes.node,
-    title: PropTypes.string,
-    description: PropTypes.string,
+    title: PropTypes.oneOfType([PropTypes.node, PropTypes.func]),
+    description: PropTypes.oneOfType([PropTypes.node, PropTypes.func]),
     handleApplyClick: PropTypes.func,
     handleCancelClick: PropTypes.func,
   }
 
   static defaultProps = {
     theme: {},
-    title: '',
-    description: '',
   }
 
   render() {
--- a/packages/react-ui-ag/src/Filters/FilterCard.js
+++ b/packages/react-ui-ag/src/Filters/FilterCard.js
@@ -1,0 +1,71 @@
+import React, { PureComponent } from 'react'
+import PropTypes from 'prop-types'
+import classnames from 'classnames'
+import themed from 'react-themed'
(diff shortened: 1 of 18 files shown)


bethstevic:
Looks good 👍

packages/react-ui-ag/src/Filters/FilterCard.js
```diff
@@ -0,0 +1,71 @@
...
+        )}
+        {...props}
+      >
+        {title && <Title className={theme.FilterCard_Title}>{title}</Title>}
+        {description && <Text className={theme.FilterCard_Description}>{description}</Text>}
+        <div className={theme.FilterCard_Body}>
+          {children}
+        </div>
+        {handleApplyClick &&
+          <Button
```
nray621: Should we be spreading props to these buttons in case there is something else that might need to get passed down to them?

nray621: Should we be giving them a `className`?
`theme.FilterCard_ApplyButton`?

packages/react-ui-ag/src/Filters/FilterCard.js
```diff
@@ -0,0 +1,71 @@
...
+@themed('*', {
+  pure: true,
+})
+
+export default class FilterCard extends PureComponent {
+  static propTypes = {
+    className: PropTypes.string,
+    theme: PropTypes.object,
+    children: PropTypes.node,
+    title: PropTypes.string,
```
tadjohnston: I think `title` and `description` should be allowed to be a `PropTypes.node` or `PropTpyes.func`.

packages/react-ui-ag/src/Filters/FilterCard.js
```diff
@@ -0,0 +1,71 @@
...
+    description: PropTypes.string,
+    handleApplyClick: PropTypes.func,
+    handleCancelClick: PropTypes.func,
+  }
+
+  static defaultProps = {
+    theme: {},
+    title: '',
```
tadjohnston: Do we even need `title` and `description` defaults?

mneumark: I guess not.

packages/react-ui-ag/src/Filters/FilterCard.js
mneumark: I usually hesitate to spread props in more than one place because where everything goes can get out of control. It IS spread to the `Card`.

PR #96, opened 2017-12-11
Nr/ya/drop down menu

yasabere:
DropDownMenu component

[Card](https://rentpath.leankit.com/card/586157673)

- Drop down Menu component


bethstevic:
Looks good! 👍

packages/react-ui-core/src/DropDownMenu/ItemWrapper.js
```diff
@@ -0,0 +1,52 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+import cn from 'classnames'
+import themed from 'react-themed'
+
+@themed('DropDownMenu', { pure: true })
```
tadjohnston: I think you need to use `@themed(/^DrownDownMenu/, { pure: true })` since you're using `DropDownMenu_ListItem` and `DropDownMenu_Selected`.

packages/react-ui-core/src/DropDownMenu/ItemWrapper.js
```diff
@@ -0,0 +1,52 @@
...
+      index,
+      theme,
+      children,
+      onSelectionClick,
+      ...props
+    } = this.props
+
+    return (
+      <div
+        data-tid={'dropdown-menu-list-item'}
```
tadjohnston: Nitpick: Can't we just write this like:

```
data-tid="dropdown-menu-list-item"
```

packages/react-ui-core/src/DropDownMenu/__tests__/DropDownMenu-test.js
```diff
@@ -0,0 +1,51 @@
...
+  DropDownMenu: 'DropDownMenu',
+  DropDownMenu_ListItem: 'DropDownMenu_ListItem',
+  DropDownMenu_Selected: 'DropDownMenu_Selected',
+}
+
+const items = [1, 2, 3]
+
+describe('DropDownMenu', () => {
+  const setup = props => {
+    const wrapper = mount(
```
tadjohnston: Do we need to `mount` here or can we use `shallow`?

packages/react-ui-core/src/DropDownMenu/__tests__/DropDownMenu-test.js
```diff
@@ -0,0 +1,51 @@
...
+      theme,
+      options: items,
+      onSelection: value => {
+        testObject.value = value
+      },
+    })
+
+    expect(testObject.value).toEqual(-1)
+    wrapper.find('[data-tid="dropdown-menu-list-item"]').at(2).simulate('mouseEnter')
+    wrapper.find('[data-tid="dropdown-menu-list-item"]').at(2).simulate('click')
```
tadjohnston: It would be better to combine these into 1 lookup.
```
const listItemTwo = wrapper.find('[data-tid="dropdown-menu-list-item"]').at(2)
listItemTwo.simulate('mouseenter')
listItemTwo.simulate('click')
```

packages/react-ui-core/src/List/List.js
```diff
@@ -70,7 +70,13 @@ export default class List extends PureComponent {
         )}
       >
         {items.map((item, i) => (
-          <Item key={this.itemId(i)} nodeType={listItemNodeType} {...props}>
+          <Item
+            key={this.itemId(i)}
+            index={i}
+            nodeType={listItemNodeType}
+            theme={theme}
```
tadjohnston: You shouldn't be passing `theme` here. It gets passed down by the theme provider through context.

stories/core/DropDownMenu.js
```diff
@@ -0,0 +1,26 @@
...
+  <DropDownMenu
+    options={['Option1', 'Option2', 'Option3']}
+  />
+)
+
+export const DropDownMenuOnSelection = (
+  <DropDownMenu
+    options={['Option1', 'Option2', 'Option3']}
+    onSelection={value => {
+      alert(value)
```
tadjohnston: Can you use the actions addon instead here?  Take a look at the Modal story for an example.  https://github.com/rentpath/react-ui/blob/master/stories/core/Modal.js#L25

stories/core/DropDownMenu.js
```diff
@@ -0,0 +1,26 @@
...
+    }}
+  />
+)
+
+export const DropDownMenuOnSelectionHover = (
+  <DropDownMenu
+    options={['Option1', 'Option2', 'Option3']}
+    onSelectionHover={value => {
+      alert(value)
```
tadjohnston: Same as above. I really hope this isn't alerting every hover 😟

stories/theme/core/DropDownMenu.css
```diff
@@ -0,0 +1,13 @@
+.DropDownMenu  {
+  padding: 0;
+  width: 100px;
+  background: grey;
+}
+
+.DropDownMenu_Selected {
+  background: lightgrey
```
tadjohnston: Need `;`

packages/react-ui-core/src/DropDownMenu/__tests__/DropDownMenu-test.js
nray621: Just doubled checked, looks like it needs to be a full `mount`

packages/react-ui-core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@rentpath/react-ui-core",
-  "version": "1.4.0",
+  "version": "1.5.0",
```
tadjohnston: Not sure what's going on with you rebase, we're already on `1.5.0` in master.

packages/react-ui-core/src/DropDown/DefaultAnchor.js
```diff
@@ -0,0 +1,48 @@
+import React, { PureComponent } from 'react'
+import PropTypes from 'prop-types'
+import { Button } from '../Button'
+
+export default class DefaultAnchor extends PureComponent {
```
tadjohnston: What do you think about renaming this to `AnchorButton` or `DropdownAnchorButton`? That would be a more descriptive name as to what this component is.  If you name is `DropdownAnchorButton` I would actually export this so the user can use if it they want.

packages/react-ui-core/src/DropDown/DropDown.js
```diff
@@ -0,0 +1,92 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+import themed from 'react-themed'
+import classnames from 'classnames'
+import DefaultAnchor from './DefaultAnchor'
+
+@themed(/^DropDown/)
+
+export default class DropDown extends Component {
```
tadjohnston: What do you think about calling this `Dropdown` instead?  This casing seems to be the norm if you do a search on the web.  

https://v4-alpha.getbootstrap.com/components/dropdowns/
https://semantic-ui.com/modules/dropdown.html
http://materializecss.com/dropdown.html

packages/react-ui-core/src/DropDown/DefaultAnchor.js
nray621: Fair enough. I'll make that change

packages/react-ui-core/src/DropDown/DropDown.js
nray621: Yea, that works. We should probably follow that convention.

packages/react-ui-core/src/DropDownMenu/ItemWrapper.js
```diff
@@ -0,0 +1,52 @@
...
+    } = this.props
+
+    return (
+      <div
+        data-tid="dropdown-menu-list-item"
+        className={cn(
+          theme.DropDownMenu_ListItem,
+          selectedIndex === index && theme.DropDownMenu_Selected
+        )}
+        onClick={onSelectionClick}
```
tadjohnston: We will most likely have an issue here.  If someone passes in `onClick` it will override this based on the spreading of props further down.  Suggestions:
* disallow `onClick` prop from being spread
* rename `onSelectionClick` to `onClick` (this would be my choice 😄 )

nray621: Gotcha. I'll go with your suggestion.

packages/react-ui-core/src/DropDownMenu/ItemWrapper.js
```diff
@@ -0,0 +1,52 @@
...
+        onMouseEnter={this.handleHover}
```
tadjohnston: I wonder if we should call `onSelectionHover` `onMouseEnter` so it explains what it's actually doing.  That would also mean you'd have to destructure it out above though.

packages/react-ui-core/src/DropDownMenu/ItemWrapper.js
tadjohnston: Whatever works, more of a suggestion.  Either way though we'd have to fix something 👍

packages/react-ui-core/src/DropDownMenu/__tests__/DropDownMenu-test.js
```diff
@@ -0,0 +1,52 @@
...
+    }
+  }
+
+  it('Should pass options to List component', () => {
+    const { wrapper } = setup({ theme, options: items })
+
+    expect(wrapper.find('[data-tid="dropdown-menu-list-item"]').length).toEqual(3)
+  })
+
+  it('Should perform on click funcationality ', () => {
```
tadjohnston: Nitpick: I don't this we capitalize the first word in the `it` statements 😄

nray621: 👍

packages/react-ui-core/src/DropDownMenu/__tests__/DropDownMenu-test.js
```diff
@@ -0,0 +1,52 @@
...
+      },
+    })
+
+    expect(testObject.value).toEqual(-1)
+    const listItemTwo = wrapper.find('[data-tid="dropdown-menu-list-item"]').at(2)
+    listItemTwo.simulate('mouseenter')
+    listItemTwo.simulate('click')
+    expect(testObject.value).toEqual(3)
+  })
+})
```
tadjohnston: Please add tests for the edge cases here: https://github.com/rentpath/react-ui/pull/96/files#diff-353e6e7bb5bbc437879857813f5532d3R84

packages/react-ui-core/src/DropDownMenu/__tests__/DropDownMenu-test.js
tadjohnston: Spelling error on `funcationality`.  Can we also be more specific on what exactly you're testing? The wording here seems like it would be more of a `describe` statement than an `it`.

packages/react-ui-core/src/DropDownMenu/ItemWrapper.js
nray621: So, I think we didn't call the prop `onClick` because we didn't want to have an `onClick` event handler for the entire `List`. So I'll go with destructuring `onClick` in render to make sure it doesn't override ours.

packages/react-ui-core/src/DropDownMenu/ItemWrapper.js
nray621: Same as above, we needed it do not have a reserved event function name (`onMouseEnter`) when it is passed to `List` and then passed on to `ItemWrapper` because we didn't want list to inadvertently have event handlers applied to the whole `List` rather than the individual items.

tadjohnston: Perfectly fine 👍

PR #97, opened 2017-12-12
Bedroom Filter Card

mneumark:
[Story](https://rentpath.leankit.com/card/583481647)

Paired with Anye

* Create BedroomFilterCard
* Filter Card changes
  * move title and description into header
  * move buttons into body
* remove handleChange from props in RadioGroup
* change name of Content className to Card_Body
* fix misuse of onChange in RadioGroup story
* add CSS to design inline card and desktop searchbar widget

### DIFF
--- a/packages/react-ui-ag/src/Buttons/ApplyButton.js
+++ b/packages/react-ui-ag/src/Buttons/ApplyButton.js
@@ -1,0 +1,49 @@
+import React, { PureComponent } from 'react'
+import PropTypes from 'prop-types'
+import classnames from 'classnames'
+import themed from 'react-themed'
+import { Button } from '@rentpath/react-ui-core'
+
+@themed(['ApplyButton'], {
+  pure: true,
+})
+
+export default class ApplyButton extends PureComponent {
+  static propTypes = {
+    className: PropTypes.string,
+    onClick: PropTypes.func,
+    theme: PropTypes.object,
+    name: PropTypes.node,
+  }
+
+  static defaultProps = {
+    theme: {},
+    name: 'Apply',
+  }
+
+  render() {
+    const {
+      theme,
+      name,
+      className,
+      onClick,
+      ...props
+    } = this.props
+
+    if (!onClick) return null
+
+    return (
+      <Button
+        className={classnames(
+          theme.ApplyButton,
+          className,
+        )}
+        onClick={onClick}
+        data-tid="apply-button"
+        {...props}
+      >
+        {name}
+      </Button>
+    )
+  }
+}
--- a/packages/react-ui-ag/src/Buttons/CancelButton.js
+++ b/packages/react-ui-ag/src/Buttons/CancelButton.js
@@ -1,0 +1,49 @@
+import React, { PureComponent } from 'react'
+import PropTypes from 'prop-types'
+import classnames from 'classnames'
+import themed from 'react-themed'
+import { Button } from '@rentpath/react-ui-core'
+
+@themed(['CancelButton'], {
+  pure: true,
+})
+
+export default class CancelButton extends PureComponent {
+  static propTypes = {
+    className: PropTypes.string,
+    onClick: PropTypes.func,
+    theme: PropTypes.object,
+    name: PropTypes.node,
+  }
+
+  static defaultProps = {
+    theme: {},
+    name: 'Cancel',
+  }
+
+  render() {
+    const {
+      theme,
+      name,
+      className,
+      onClick,
+      ...props
+    } = this.props
+
+    if (!onClick) return null
+
+    return (
+      <Button
+        className={classnames(
+          theme.CancelButton,
+          className,
+        )}
+        onClick={onClick}
+        data-tid="cancel-button"
+        {...props}
+      >
+        {name}
+      </Button>
+    )
+  }
+}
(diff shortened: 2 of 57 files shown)


bethstevic:
Looks good to me 👍

packages/react-ui-ag/src/Filters/BedroomFilterCard.js
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  constructor(props) {
+    super(props)
+    this.state = {
+      value: (props.fields.find(f => f.checked) || {}).value,
+    }
+  }
+
+  handleRadioGroupSelection = event => {
```
tadjohnston: So did we decide to use this style in favor of the `@autobind` decorator?

packages/react-ui-ag/src/Filters/BedroomFilterCard.js
```diff
@@ -0,0 +1,91 @@
...
+      { label: '2', value: '2' },
+      { label: '3', value: '3' },
+      { label: '4+', value: '4' },
+    ],
+  }
+
+  constructor(props) {
+    super(props)
+    this.state = {
+      value: (props.fields.find(f => f.checked) || {}).value,
```
tadjohnston: What happens if `prop.fields` happens to change after component mount 😃 ?

stories/theme/ag/FilterCard.css
```diff
@@ -12,3 +12,183 @@
   padding: 25px;
   margin-bottom: 5px;
 }
+
+.InlineCard .RadioButton,
```
tadjohnston: Isn't there a generic class you can use between the InlineCard and DesktopCard?

stories/theme/ag/FilterCard.css
```diff
@@ -12,3 +12,183 @@
...
+.DesktopCard .RadioButton {
+  display: inline-block;
+  padding: 0;
+}
+
+.InlineCard .FilterCard_Body,
```
tadjohnston: Same as above.

stories/theme/ag/FilterCard.css
```diff
@@ -12,3 +12,183 @@
...
+  padding: 0;
+}
+
+/***********************/
+/* Inline Card Styling */
+/***********************/
+
+.InlineCard {
+  background-color: #0175d7;
+  font-family: Helvetica;
```
tadjohnston: You can remove this now since we'll have a fancy font from my PR 😄

packages/react-ui-ag/src/Buttons/ApplyButton.js
```diff
@@ -0,0 +1,49 @@
+import React, { PureComponent } from 'react'
+import PropTypes from 'prop-types'
+import classnames from 'classnames'
+import themed from 'react-themed'
+import { Button } from '@rentpath/react-ui-core'
+
+@themed(/^ApplyButton/, {
```
tadjohnston: If you're not using any other classes like `ApplyButton-foo` then you can just do:
```
@themed(['ApplyButton'], {
```

packages/react-ui-ag/src/Buttons/ApplyButton.js
```diff
@@ -0,0 +1,49 @@
...
+
+    return (
+      <Button
+        className={classnames(
+          theme.ApplyButton,
+          className,
+        )}
+        onClick={handleClick}
+        data-tid="apply-button"
+        {...props}
```
tadjohnston: Any `onClick` passed will override your `handleClick` because of this.  Is that what you want?

packages/react-ui-ag/src/Buttons/CancelButton.js
```diff
@@ -0,0 +1,49 @@
...
+
+    return (
+      <Button
+        className={classnames(
+          theme.CancelButton,
+          className,
+        )}
+        onClick={handleClick}
+        data-tid="cancel-button"
+        {...props}
```
tadjohnston: Same thing as [above](https://github.com/rentpath/react-ui/pull/97/files#r156829655)

packages/react-ui-ag/src/Filters/BedroomFilterCard.js
```diff
@@ -0,0 +1,136 @@
...
+export default class BedroomFilterCard extends Component {
+  static propTypes = {
+    className: PropTypes.string,
+    theme: PropTypes.object,
+    title: PropTypes.string,
+    description: PropTypes.string,
+    applyButtonProps: PropTypes.shape({
+      handleClick: PropTypes.func,
+    }),
+    cancelButtonProps: PropTypes.shape({
```
tadjohnston: Instead of this you _could_ export the ApplyButton and Cancel button and allow the user to pass the component back in with the extra props attached.  That way they still get the handleClick functionality and separations of concerns is better IMO.

packages/react-ui-ag/src/Buttons/ApplyButton.js
mneumark: I think you don't need the array IIRC. :+1:

packages/react-ui-ag/src/Buttons/ApplyButton.js
mneumark: Probably not. I'll move it. Good call.

PR #98, opened 2017-12-13
misc fixes and storybook cleanup

tadjohnston:
[Story](https://rentpath.leankit.com/card/588092909)

* cleaned up storybook to work better with themed components
* when cleaning up found bugs in code that needed to be fixed
  * context theme bug fix for modal
  * className bug fix for ratingbar
  * added @lfender6445 fix from https://github.com/rentpath/react-ui/pull/93 in here as well.

PR #99, opened 2017-12-13
Fix double onClick handler on Counter

tcwilson:
also fixes lint and tests

counter-bug


lfender6445:
This was actually fixed here tim https://github.com/rentpath/react-ui/pull/98

PR #100, opened 2017-12-13
Allow onChange in RequiredField

orionstein:
[577966483](https://rentpath.leankit.com/card/577966483)

Provides support for additional onChange effects, needed for storing input values used in a RequiredField in the leads cookie


packages/react-ui-rent/src/LeadForm/Fields/RequiredField.js
```diff
@@ -44,6 +45,10 @@ export default class RequiredField extends Component {
     } else {
       this.setState({ labelVisible: true })
     }
+    if (this.props.onChange &&
+      typeof this.props.onChange === 'function') {
```
tcwilson: probably don't need the typeof check, as you have the prop type validation above

tcwilson:
👍

PR #102, opened 2017-12-13
dropdown items tweaks and renames

tadjohnston:
[Story](https://rentpath.leankit.com/card/589072251)

* changing item names to something more contextual
* adding default for Item
* fixing lint and test errors
* adding tests
* changing theme class names
* added support (lower cased props) for any prop that takes in a component for function

Worked with @nray621 to address the issues.


packages/react-ui-core/src/Dropdown/Dropdown.js
```diff
@@ -79,12 +80,11 @@ export default class Dropdown extends Component {
           )}
         />
         {dropDownVisible &&
-          <div
+          <Card
             data-tid="dropdown-content"
```
mneumark: We seem to be mixed when it comes to naming this kind of thing. In some places its called Body, in others its called Content. I think most its Body, can we stay consistent?

packages/react-ui-core/src/Dropdown/DropdownMenu.js
```diff
@@ -90,29 +93,25 @@ export default class DropdownMenu extends Component {
...
         )}
         items={options}
-        Item={ItemWrapper}
-        nodeType="div"
-        selectedIndex={activeIndex}
-        handleSelectionClick={this.handleSelection}
-        handleSelectionHover={this.activateOption}
+        activeIndex={this.state.activeIndex}
+        onClick={this.handleSelection}
+        onMouseEnter={this.activateOption}
```
mneumark: What does activateOption mean? Is that highlighting an item? If so that could be clearer...

packages/react-ui-core/src/Dropdown/__tests__/DropdownMenu-test.js
```diff
@@ -26,8 +26,7 @@ describe('DropdownMenu', () => {
 
   it('should pass options to List component', () => {
     const { wrapper } = setup({ theme, options: items })
-
-    expect(wrapper.find('[data-tid="dropdown-menu-list-item"]').length).toEqual(3)
+    expect(wrapper.find('ListItem')).toHaveLength(3)
```
mneumark: 👍

packages/react-ui-core/src/List/ListItem.js
```diff
@@ -12,18 +15,32 @@ export default class ListItem extends PureComponent {
...
+    index: PropTypes.number,
+    active: PropTypes.bool,
   }
 
   static defaultProps = {
     theme: {},
     nodeType: 'li',
   }
 
+  @autobind
```
mneumark: Awesome!

packages/react-ui-core/src/Dropdown/Dropdown.js
tadjohnston: Sure, whatever we cant to call it is fine.  I'll change it to body.

packages/react-ui-core/src/Dropdown/DropdownMenu.js
tadjohnston: It really could mean anything I think, that's why it's vague. It generally would mean hovered or selected with up / down keyboard stroke, but it's not _technically_ selected yet. Make sense yet? ;)

mneumark: thats why I said highlighting because that seems to make it more clear while leaving any opening for it to be a little vague.

tadjohnston: Ok, I can change it.

packages/react-ui-core/src/Dropdown/DropdownMenu.js
```diff
@@ -46,30 +50,36 @@ export default class DropdownMenu extends Component {
     document.addEventListener('keydown', this.onKeyDown)
```
grenti: should this be in componentDidMount()?

tadjohnston: This was there before, but it looks like it yeah. Where do you think it should go?

grenti: I've never used it in constructor only after the component was mounted in componentDidMount() - there may be a valid reason to do so so though that's why I was asking.

nray621: I think it is better suited for `componentDidMount` as well.

nray621: Not sure who put it in `constructor` (decent chance it was me 😉 ) My bad if so

tadjohnston: Am I missing something? I haven't changed this and it's in `componentDidMount`. Did you expand the code?

nray621: Nope, you aren't missing anything... it is in `componentDidMount`. Should have double checked. What are you talking about @grenti ??

grenti: Sorry about that - I thought I saw it in the constructor. LGTM 👍

grenti:
Looks good to me 👍

packages/react-ui-core/src/List/List.js
```diff
@@ -71,9 +76,11 @@ export default class List extends PureComponent {
       >
         {items.map((item, i) => (
           <Item
+            {...itemProps}
+            highlight={!!(highlightIndex === i)}
```
nray621: `highlightIndex === i` isn't already a bool? I'm guessing that's why you're doing `!!`

tadjohnston: Yeah, I added this for some reason before, I don't think we need it now.  It may have been for some test purpose.

packages/react-ui-core/src/Dropdown/Dropdown.js
```diff
@@ -66,25 +73,20 @@ export default class Dropdown extends Component {
           theme.Dropdown,
           className
         )}
+        {...props}
       >
-        <Anchor
-          data-tid="dropdown-anchor"
-          {...props}
+        <Field
```
nray621: Why are we changing the default anchor to a field? I feel like a button makes more sense as the default.

tadjohnston: Well.. the thought was for the search bar, but yeah I think you're right, let me change it back.

PR #103, opened 2017-12-14
Auto suggest field

nray621:
[story](https://rentpath.leankit.com/card/586157770)

Create `AutoSuggestField` composed of `Dropdown` and `DropdownMenu` as well as optional clear and submit `Button`s


tadjohnston:
This shouldn't be blocked anymore. If it still is, please let me know.

PR #104, opened 2017-12-14
Price Filter Component

bethstevic:
[Card](https://rentpath.leankit.com/card/583480451)

- Create a generic component for the price filter


packages/react-ui-ag/src/Filters/PriceFilterCard.js
```diff
@@ -0,0 +1,93 @@
...
+  handleCancelClick() {
+    const { onCancelClick } = this.props
+
+    if (onCancelClick) onCancelClick(this.state.value)
+  }
+
+  renderPriceSlider() {
+    const { priceSlider } = this.props
+    const [Slider, props] = parseArgs(priceSlider, RangeSlider)
+    return (<Slider {...(props || {})} onChangeComplete={this.handlePriceSliderChange} />)
```
mneumark: Can you experiment with removing the parentheses and the `(props || {})`? I think in the end they weren't needed.

tadjohnston:
@bethstevic mind fixing the commented out code so we can review this?

bethstevic:
@tadjohnston that was leftover from Micah and Anye's PR when they moved the file structure around. I am nervous about deleting it, but, if you and Micah think it is definitely no longer needed I can do that

tadjohnston:
We need to do something with it. I'm surprised this branch doesn't have conflicts. The package.json file is wrong in this PR as well for `react-ui-core`. master is on `1.5.0`.

packages/react-ui-ag/src/Filters/__tests__/PriceFilterCard-test.js
```diff
@@ -0,0 +1,69 @@
...
+    const wrapper = mount(<PriceFilterCard />, { lifecycleExperimental: true })
+    const defaultPriceSliderProps = {
+      formatLabel: val => `$${val}`,
+      value: { min: 500, max: 5000 },
+      minValue: 0,
+      maxValue: 15000,
+      step: 100,
+    }
+
+    expect(wrapper.state().hasPriceChanged).toBeFalsy()
```
tadjohnston: You could also set a const here so it doesn't have to look up each time:
```
const { priceSlider, hasPriceChanged  = wrapper.state()
// or even
// const priceSlider = const wrapper.state('priceSlider')

expect(hasPriceChanged).toBeFalsy()
expect(priceSlider.minValue).toEqual(defaultPriceSliderProps.minValue)
expect(priceSlider.maxValue).toEqual(defaultPriceSliderProps.maxValue)
expect(priceSlider.step).toEqual(defaultPriceSliderProps.step)
expect(priceSlider.value).toEqual(defaultPriceSliderProps.value)
```

PR #108, opened 2017-12-16
Rating Filter Card

mneumark:
[Story](https://rentpath.leankit.com/card/589790911)

* created RatingFilterCard
* created storybook example showing inline card (no desktop equivalent)

### DIFF
--- a/packages/react-ui-ag/src/Filters/RatingFilterCard.js
+++ b/packages/react-ui-ag/src/Filters/RatingFilterCard.js
@@ -1,0 +1,91 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+import classnames from 'classnames'
+import themed from 'react-themed'
+import isEqual from 'lodash/isEqual'
+import { RadioGroup } from '@rentpath/react-ui-core'
+import autobind from 'autobind-decorator'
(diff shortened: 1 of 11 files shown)


packages/react-ui-ag/src/Filters/RatingFilterCard.js
```diff
@@ -0,0 +1,91 @@
...
+    className: PropTypes.string,
+    theme: PropTypes.object,
+    fields: PropTypes.arrayOf(PropTypes.shape({
+      label: PropTypes.string,
+      value: PropTypes.string,
+    })),
+  }
+
+  static defaultProps = {
+    title: 'View 4+ stars units in Atlanta',
```
grenti: Just a slight nit-pick - I'm ok if there is consensus not to change it, but maybe we should take off Atlanta as a default? Probably just `View 4+ star units'?

grenti:
LGTM 👍

packages/react-ui-ag/src/Filters/RatingFilterCard.js
mneumark: Doesn't matter. I left it there so one can see what it should look like but in the real app I think ag.js would pass the title dynamically.

PR #109, opened 2017-12-16
Ensure label disappears on default value

tcwilson:


orionstein:
👍 If we use RequiredLabel for a controlled input, will be needed

paulpostel:
👍

PR #110, opened 2017-12-16
auto suggest field

nray621:
[story](https://rentpath.leankit.com/card/586157770)

Create `AutoSuggestField` composed of `Dropdown` and `DropdownMenu` as well as optional clear and submit `Button`s


packages/react-ui-core/src/AutoSuggestField/InputAnchor.js
```diff
@@ -0,0 +1,64 @@
+import React, { PureComponent } from 'react'
```
tadjohnston: This file does exactly what the [AnchorButton](https://github.com/rentpath/react-ui/blob/master/packages/react-ui-core/src/Dropdown/AnchorButton.js) does.  Can we please try to use that component and modify it so it will work for both an input and button?

nray621: Fair enough. So give `AnchorButton` a prop to be able to be a `Field` optionally?

packages/react-ui-core/src/AutoSuggestField/AutoSuggestField.js
```diff
@@ -0,0 +1,205 @@
...
+  @autobind
+  handleVisibilityChange(visible) {
+    this.setState({
+      visible,
+    })
+  }
+
+  @autobind
+  reset() {
+    this.setState({ value: '', visible: false })
```
tadjohnston: Why is the value of `visible` here different than in the constructor? Also, the value of `state.value` is set to the prop in the constructor.  How does this reset if the original value is say `foo`?

packages/react-ui-core/src/AutoSuggestField/AutoSuggestField.js
```diff
@@ -0,0 +1,205 @@
...
+    onAfterReset: () => {},
+    onSubmit: () => {},
+    onInput: () => {},
+    onSelection: () => {},
+    submitOnSelection: true,
+  }
+
+  constructor(props) {
+    super(props)
+    const { value } = this.props
```
tadjohnston: Use the props coming in, you don't need `this.props`.
```
props.value

this.onInput = debounce(props.onInput, 300, { leading: false, trailing: true })

```

packages/react-ui-core/src/AutoSuggestField/AutoSuggestField.js
```diff
@@ -0,0 +1,205 @@
...
+    placeholder: PropTypes.string,
+    suggestionValueSelector: PropTypes.func,
+    onSelectionHover: PropTypes.func,
+    submitOnSelection: PropTypes.bool,
+  }
+
+  static defaultProps = {
+    theme: {},
+    value: '',
+    onAfterReset: () => {},
```
tadjohnston: Do we really need to default these or can we just check for the existence and execute based on that instead?

packages/react-ui-core/src/AutoSuggestField/AutoSuggestField.js
```diff
@@ -0,0 +1,205 @@
...
+  }
+
+  @autobind
+  reset() {
+    this.setState({ value: '', visible: false })
+    this.props.onAfterReset()
+  }
+
+  renderButton(button, handleClick, dataTid, props) {
+    if (!button) return undefined
```
tadjohnston: `null` is the appropriate return value here.  React needs either a node or null to render with.

packages/react-ui-core/src/AutoSuggestField/AutoSuggestField.js
```diff
@@ -0,0 +1,205 @@
...
+          <Menu
+            options={suggestions}
+            onSelection={this.handleSuggestionSelection}
+            handleSelectionHover={onSelectionHover}
+            optionValueSelector={suggestionValueSelector}
+          />
+        </Dropdown>
+
+        {this.renderButton(submitButton, this.handleOnSubmit, 'autosuggest-submit-button', ...props)}
+        {this.renderButton(resetButton, this.reset, 'autosuggest-reset-button', ...props)}
```
tadjohnston: How about something more like this:
```
this.renderButton(resetButton, { ...props, onClick: this.reset, 'data-tid': 'autosuggest-reset-button' })
```

This way you only have 2 arguments to worry about.

```
renderButton(button, props) {
  if (!button) return null

  return createElement(...parseArgs(button, Button, props))
}
```

Please also be careful about passing props like this here.  This means that both buttons get the same treatment as the `Dropdown`.

packages/react-ui-core/src/AutoSuggestField/InputAnchor.js
```diff
@@ -0,0 +1,49 @@
...
+
+  render() {
+    const {
+      handleDocumentClick,
+      changeVisibility,
+      toggleVisibilty,
+      visible,
+      theme,
+      text,
+      onClick, // eslint-disable-line no-unused-vars
```
tadjohnston: This is used so not sure why the `eslint` disable line is here.

packages/react-ui-core/src/AutoSuggestField/AutoSuggestField.js
tadjohnston: Hmm.. I guess they've since changed this: https://reactjs.org/docs/jsx-in-depth.html#booleans-null-and-undefined-are-ignored

`undefined` is fine here.

packages/react-ui-core/src/AutoSuggestField/AutoSuggestField.js
yasabere: At the moment the suggestions list is closed on reset by default but I think it definitely makes sense to have that be configurable through a prop

packages/react-ui-core/src/AutoSuggestField/AutoSuggestField.js
yasabere: This has been used in other places in the codebase

packages/react-ui-core/src/AutoSuggestField/AutoSuggestField.js
yasabere: 👍

packages/react-ui-core/src/AutoSuggestField/AutoSuggestField.js
tadjohnston: Well, in some cases I would say it's fine.  Also, just because it's in other places in code doesn't mean those places are correct 😄

packages/react-ui-core/src/AutoSuggestField/AutoSuggestField.js
```diff
@@ -0,0 +1,216 @@
...
+      theme,
+      clearButton,
+      submitButton,
+      anchorField,
+      suggestions,
+      className,
+      onSubmit, // eslint-disable-line no-unused-vars
+      onItemMouseOver,
+      onSelection, // eslint-disable-line no-unused-vars
+      valueSelector,
```
bethstevic: Why is onSubmit and onSelection not used?

tadjohnston: They aren't used here, but they are used in `handleSelectionSuggestion` and `handleSubmit` so they shouldn't be passed on to the `Dropdown`.

nray621: We need to make sure they don't get passed down through spread props, so we're pulling them out even though we aren't using them in this scope.

bethstevic: Ah, that makes sense. You are pulling them out so they don't get passed down.

bethstevic:
Looks good to me 👍

vamshi100:
Looks good to me 👍

stories/core/Dropdown.js
```diff
@@ -55,14 +61,15 @@ export const DropdownWithMenu = (
     className={classnames(
       StoryBookTheme['Story-padding'],
       StoryBookTheme.Story_Menu,
+      theme.multiple
```
tadjohnston: Let's keep the casing standard everywhere.

stories/theme/core/Dropdown.css
```diff
@@ -36,5 +37,17 @@
 
 .Dropdown .Button-collapse:after {
   content: '\25bc';
+}
+
+.Dropdown .List {
+  padding: 5px;
+}
 
+.multiple .Text {
```
tadjohnston: If this is specific to a class outside of core, I'd much rather prefer it in the storybook theme.

PR #111, opened 2017-12-19
Add closeOnSelect and toggleVisibility props to Menu

nray621:
[story](https://rentpath.leankit.com/card/591017663)

Worked with @yasabere and @mneumark 

Allow Dropdown's children to control visibility through `toggleVisibility` function.
Add `toggleVisibility` and `closeOnSelection`(bool) props to `Menu`
Change `visible: this.nextProps.visible` to `visible: nextProps.visible`

### DIFF
--- a/packages/react-ui-core/src/Dropdown/Dropdown.js
+++ b/packages/react-ui-core/src/Dropdown/Dropdown.js
@@ -49,16 +49,23 @@ export default class Dropdown extends Component {
     }
   }
 
-  renderAnchor(anchorField) {
-    return parseArgs(anchorField, AnchorButton)
+  renderAnchor() {
+    const [Anchor, fieldProps] = parseArgs(this.props.anchorField, AnchorButton, {
+      'data-tid': 'dropdown-anchor',
+      visible: this.state.visible,
+      handleDocumentClick: this.handleDocumentClick,
+      toggleVisibilty: this.toggleVisibilty,
+    })
+
+    return <Anchor {...fieldProps} />
   }
 
   renderChildren() {
     const props = { toggleVisibilty: this.toggleVisibilty }
     const children = React.Children.toArray(this.props.children)
-    return React.Children.map(children, child =>
-      (typeof child.type === 'function' ? cloneElement(child, props) : child)
-    )
+    return React.Children.map(children, child => (
+      typeof child.type === 'function' ? cloneElement(child, props) : child
+    ))
   }
 
   render() {
@@ -71,8 +78,6 @@ export default class Dropdown extends Component {
       ...props
     } = this.props
 
-    const [Anchor, fieldProps] = this.renderAnchor(anchorField)
-    const dropDownVisible = this.state.visible
     return (
       <div
         ref={ref => { this.dropdown = ref }}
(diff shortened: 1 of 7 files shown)


packages/react-ui-core/src/Dropdown/Dropdown.js
```diff
@@ -53,6 +53,16 @@ export default class Dropdown extends Component {
     return parseArgs(anchorField, AnchorButton)
   }
 
+  renderChildren() {
+    const {
+      children,
+    } = this.props
+
+    return cloneElement(children, {
```
tadjohnston: What if children is an array of children?

tadjohnston:
Did you add tests for this code? 😄

nray621:
@tadjohnston Will do. Was just trying to get the blocker out of the way for Micah and Rent people

packages/react-ui-core/src/Dropdown/Dropdown.js
nray621: Good point. I'll use `React.children` and iterate over them if that is the case

tadjohnston: You probably want: 
```
const childrenWithProps = React.Children.map(children, child =>
  React.cloneElement(child, { toggleVisibility: this.toggleVisibility }))

return childrenWithProps
```

nray621: haha I literally just wrote exactly that 😉

packages/react-ui-core/src/Dropdown/Dropdown.js
```diff
@@ -53,6 +53,25 @@ export default class Dropdown extends Component {
     return parseArgs(anchorField, AnchorButton)
   }
 
+  renderChildren() {
+    const {
+      children,
+    } = this.props
+
+    if (Array.isArray(children)) {
+      const clonedChildren = React.Children.map(children, child =>
```
tadjohnston: Why don't you just return instead of setting the const?

packages/react-ui-core/src/Dropdown/Dropdown.js
```diff
@@ -53,6 +53,25 @@ export default class Dropdown extends Component {
...
+        cloneElement(child, {
+          toggleVisibilty: this.toggleVisibilty,
+        })
+      )
+      return clonedChildren
+    }
+
+    return cloneElement(children, {
+      toggleVisibilty: this.toggleVisibilty,
```
tadjohnston: If you're passing the props in two possible ways, maybe it would be better to set this as a const at the top and use it?
```
const props = { toggleVisibilty: this.toggleVisibilty }


if (Array.isArray(children)) {
  return React.Children.map(children, child => cloneElement(child, props))
}

return cloneElement(children, props)

File: packages/react-ui-core/src/Dropdown/Dropdown.js [hunk 4]
```diff
@@ -53,6 +53,16 @@ export default class Dropdown extends Component {
     return parseArgs(anchorField, AnchorButton)
   }
 
+  renderChildren() {
+    const props = { toggleVisibilty: this.toggleVisibilty }
+    const children = React.Children.toArray(this.props.children)
+    /* eslint-disable no-confusing-arrow */
```
tadjohnston: I'd rather not do this if we don't have to. Can't we just add the parens to the following line instead?

nray621: 👍

PR #113, opened 2017-12-26
Add highlighter

yasabere:
[card](https://rentpath.leankit.com/card/587790554)

component which adds class to text based on pattern prop

### DIFF
--- a/packages/react-ui-core/src/AutoSuggestField/AutoSuggestField.js
+++ b/packages/react-ui-core/src/AutoSuggestField/AutoSuggestField.js
@@ -7,6 +7,7 @@ import autobind from 'autobind-decorator'
 import themed from 'react-themed'
 import classnames from 'classnames'
 import { Dropdown } from '../Dropdown'
+import { Highlighter } from '../Highlighter'
 import { Menu } from '../Menu'
 import { Button } from '../Button'
 import { Field } from '../Form'
@@ -33,6 +34,10 @@ export default class AutoSuggestField extends Component {
       PropTypes.func,
       PropTypes.object,
     ]),
+    highlight: PropTypes.oneOfType([
+      PropTypes.bool,
+      PropTypes.object,
+    ]),
     onAfterClear: PropTypes.func,
     onSubmit: PropTypes.func,
     onSelection: PropTypes.func,
@@ -55,6 +60,7 @@ export default class AutoSuggestField extends Component {
     valueSelector: value => value,
     submitOnSelection: true,
     visible: false,
+    highlight: false,
   }
 
   constructor(props) {
@@ -92,6 +98,21 @@ export default class AutoSuggestField extends Component {
         }
         break
       default:
+    }
+  }
+
+  get highlightedListItem() {
+    const { highlight } = this.props
+
+    if (!highlight) return { nodeType: 'div' }
+    const indexHighlighted = highlight.indexHighlighted >= 0 ? highlight.indexHighlighted : 1
+
+    return { nodeType: props =>
+      (<Highlighter
+        indexHighlighted={indexHighlighted}
+        {...props}
+        pattern={this.state.value}
+      />),
     }
   }
 
@@ -180,6 +201,7 @@ export default class AutoSuggestField extends Component {
       onSubmit, // eslint-disable-line no-unused-vars
       onItemMouseOver,
       onSelection, // eslint-disable-line no-unused-vars
+      highlight,
       valueSelector,
       value,
       onAfterClear,
@@ -204,6 +226,7 @@ export default class AutoSuggestField extends Component {
           {...props}
         >
           <Menu
+            listItem={this.highlightedListItem}
             options={suggestions}
             onItemSelect={this.handleSuggestionSelection}
             onItemMouseOver={onItemMouseOver}
--- a/packages/react-ui-core/src/Highlighter/Highlighter.js
+++ b/packages/react-ui-core/src/Highlighter/Highlighter.js
@@ -1,0 +1,87 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+import themed from 'react-themed'
+import classnames from 'classnames'
+import { randomId } from '@rentpath/react-ui-utils'
+
+@themed(/^Highlighter/)
+
+export default class Highlighter extends Component {
+  static propTypes = {
+    className: PropTypes.string,
+    pattern: PropTypes.string.isRequired,
+    children: PropTypes.string,
+    theme: PropTypes.object,
+    indexHighlighted: PropTypes.number,
+  }
+
+  static defaultProps = {
+    text: '',
+    children: '',
+    theme: {},
+  }
+
+  get indexHighlighted() {
+    return this.props.indexHighlighted + 1
+  }
+
+  renderHighlightedText() {
+    const { pattern, children } = this.props
+
+    if (!pattern) return children
+
+    const segments = children.split(pattern)
+
+    return segments.reduce((prev, curr, index) => [prev,
+      this.renderHighlightedSegment(index),
+      curr])
+  }
+
+  renderHighlightedSegment(index) {
+    const {
+      pattern,
+      theme,
+      indexHighlighted,
+      className,
+    } = this.props
+
+    if (!indexHighlighted || index === indexHighlighted) {
+      return (
+        <span
+          data-tid={`highlighter-match-${index}`}
+          className={classnames(
+            theme['Highlighter-Match'],
+            className,
+          )}
+          key={randomId()}
+        >
+          {pattern}
+        </span>
+      )
+    }
+    return pattern
+  }
+
+  render() {
+    const {
+      theme,
+      className,
+      indexHighlighted,
+      pattern,
+      ...props
+    } = this.props
+
+    return (
+      <div
+        className={classnames(
+          theme.Highlighter,
+          className,
+        )}
+        {...props}
+      >
+        {this.renderHighlightedText()}
+      </div>
+    )
+  }
+
+}
--- a/packages/react-ui-core/src/Highlighter/__tests__/Highlighter-test.js
+++ b/packages/react-ui-core/src/Highlighter/__tests__/Highlighter-test.js
@@ -1,0 +1,44 @@
+import React from 'react'
+import { mount } from 'enzyme'
+import ThemedHighlighter from '../Highlighter'
+
+const Highlighter = ThemedHighlighter.WrappedComponent
+
+describe('Highlighter', () => {
+  it('surounds pattern text with span and classname', () => {
+    const wrapper = mount(
+      <Highlighter pattern="test">
+        This is a test sentence for a test
+      </Highlighter>
+    )
+
+    wrapper.setProps({ pattern: 'test' })
+
+    const highlighterMatch = wrapper.find('[data-tid="highlighter-match-1"]')
+    expect(highlighterMatch.length).toEqual(1)
+    const match = highlighterMatch.at(0)
+    expect(match.text()).toBe('test')
+
+    const highlighterMatch2 = wrapper.find('[data-tid="highlighter-match-2"]')
+    expect(highlighterMatch2.length).toEqual(1)
+    const match2 = highlighterMatch2.at(0)
+    expect(match2.text()).toBe('test')
+
+    wrapper.setProps({ pattern: 'test2' })
+    expect(wrapper.find('[data-tid="highlighter-match-1"]')).toHaveLength(0)
+  })
+
+  it('surounds the nth pattern matched text with span and classname', () => {
+    const wrapper = mount(
+      <Highlighter pattern="test" indexHighlighted={2}>
+        This is a test sentence for a test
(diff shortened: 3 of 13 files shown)


packages/react-ui-core/src/Highlighter/Highlighter.js
```diff
@@ -0,0 +1,66 @@
...
+      curr])
+  }
+
+  renderHighlightedSegment(index) {
+    const { pattern, theme, indexHighlighted } = this.props
+
+    if (indexHighlighted === undefined || index === indexHighlighted) {
+      return (
+        <span
+          data-tid="Highlighter-match"
```
tadjohnston: Always has to be lower cased 😃

packages/react-ui-core/src/Highlighter/Highlighter.js
```diff
@@ -0,0 +1,66 @@
...
+
+    return segments.reduce((prev, curr, index) => [prev,
+      this.renderHighlightedSegment(index),
+      curr])
+  }
+
+  renderHighlightedSegment(index) {
+    const { pattern, theme, indexHighlighted } = this.props
+
+    if (indexHighlighted === undefined || index === indexHighlighted) {
```
tadjohnston: What happens if `indexHighlighted` is null?

packages/react-ui-core/src/Highlighter/Highlighter.js
```diff
@@ -0,0 +1,66 @@
...
+    }
+    return pattern
+  }
+
+  render() {
+    const { theme } = this.props
+
+    return (
+      <div
+        className={classnames(theme.Highlighter)}
```
tadjohnston: No need for `classnames` here as you only have one class. I'm guessing you're missing `className` prop here which means you'd need `classnames` after all 😄

tadjohnston:
Let's make sure we get some description in the card so we know what this card is really supposed to be doing.

packages/react-ui-core/src/Highlighter/Highlighter.js
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  get indexHighlighted() {
+    return this.props.indexHighlighted + 1
+  }
+
+  renderHighlightedText() {
+    const { pattern, children } = this.props
+
+    const segments = children.split(pattern)
```
tadjohnston: Should this return if no pattern exists? If we're forcing this to have a pattern, then it should be required in props.

packages/react-ui-core/src/Highlighter/Highlighter.js
nray621: The propTypes enforce it as a number, so as far as I understand, it could only be `undefined` or a number at this point.

packages/react-ui-core/src/Highlighter/Highlighter.js
nray621: 👍

packages/react-ui-core/src/Highlighter/__tests__/Highlighter-test.js
```diff
@@ -0,0 +1,33 @@
...
+describe('Highlighter', () => {
+  it('surounds pattern text with span and classname', () => {
+    const wrapper = mount(
+      <Highlighter pattern="test">
+        This is a test sentence for a test
+      </Highlighter>
+    )
+
+    wrapper.setProps({ pattern: 'test' })
+    expect(wrapper.find('[data-tid="highlighter-match"]')).toHaveLength(2)
```
tadjohnston: Let's set a const to this value `wrapper.find('[data-tid="highlighter-match"]')` instead of looking it up multiple times.  It also makes it easier to change if for some reason the data-tid changes.

stories/core/Highlighter.js
```diff
@@ -0,0 +1,30 @@
...
+    <Highlighter
+      className={StoryBookTheme['Story-padding']}
+      pattern="test"
+    >
+      this is a test to see if this works, test seems to have passed the test
+    </Highlighter>
+  </div>
+)
+
+export const HighlighterWithIndex = () => {
```
tadjohnston: This story is broken.

packages/react-ui-core/src/Highlighter/__tests__/Highlighter-test.js
nray621: Sounds good 👍

packages/react-ui-core/src/Highlighter/__tests__/Highlighter-test.js
```diff
@@ -0,0 +1,33 @@
...
+    const wrapper = mount(
+      <Highlighter pattern="test" indexHighlighted={2}>
+        This is a test sentence for a test
+      </Highlighter>
+    )
+
+    wrapper.setProps({ pattern: 'test' })
+    expect(wrapper.find('[data-tid="highlighter-match"]')).toHaveLength(1)
+    wrapper.setProps({ pattern: 'test2' })
+    expect(wrapper.find('[data-tid="highlighter-match"]')).toHaveLength(0)
```
nray621: I set a const for  `wrapper.find('[data-tid="highlighter-match"]')`, but had to keep line 19 as is. I think it needs to run `find` again after the prop change on line 18

packages/react-ui-core/src/Highlighter/Highlighter.js
tadjohnston: If that's the case, can you add a test for `null`?

packages/react-ui-core/src/Highlighter/__tests__/Highlighter-test.js
tadjohnston: Most likely, yes.

stories/core/Highlighter.js
nray621: Let me take a look at why the knobs aren't working

nray621: Fixed now

packages/react-ui-core/src/Highlighter/Highlighter.js
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  get indexHighlighted() {
+    return this.props.indexHighlighted + 1
+  }
+
+  renderHighlightedText() {
+    const { pattern, children } = this.props
+
+    if (pattern === '') return children
```
tadjohnston: I think you just want `if (!pattern) return children` yes?

nray621: duh. whoops

packages/react-ui-core/src/Highlighter/Highlighter.js
```diff
@@ -0,0 +1,86 @@
...
+    }
+    return pattern
+  }
+
+  render() {
+    const {
+      theme,
+      className,
+      indexHighlighted,
+      ...props
```
tadjohnston: `pattern` needs to be destuctured here as well I believe.

packages/react-ui-core/src/Highlighter/__tests__/Highlighter-test.js
```diff
@@ -0,0 +1,34 @@
+import React from 'react'
+import { mount } from 'enzyme'
+import Highlighter from '../Highlighter'
+
+describe('Highlighter', () => {
+  it('surounds pattern text with span and classname', () => {
+    const wrapper = mount(
```
tadjohnston: I don't think we need to `mount` do we?

packages/react-ui-core/src/Highlighter/Highlighter.js
nray621: 👍

packages/react-ui-core/src/Highlighter/__tests__/Highlighter-test.js
nray621: We'll check

packages/react-ui-core/src/AutoSuggestField/AutoSuggestField.js
```diff
@@ -99,12 +102,17 @@ export default class AutoSuggestField extends Component {
   }
 
   get highlightedListItem() {
-    if (!this.props.highlight) return { nodeType: 'div' }
+    const { highlight } = this.props
+
+    if (!highlight) return { nodeType: 'div' }
+    let indexHighlighted = highlight.indexHighlighted || 1
+
+    if (highlight.indexHighlighted === 0) indexHighlighted = 0
```
tadjohnston: Why not just:
```
get highlightedListItem() {
  const { highlight } = this.props 

  if (!highlight) return { nodeType: 'div' }

  const { indexHighlighted: index, ...highlightProps } = highlight
  const indexHighlighted = index >= 0 ? index : 1

  return { 
    nodeType: props => (
      <Highlighter
        indexHighlighted={indexHighlighted}
        {...props} // not sure what this is btw
        {...highlightProps}
        pattern={this.state.value}
      />
    ),
  }
}

File: packages/react-ui-core/src/AutoSuggestField/AutoSuggestField.js [hunk 12]
```diff
@@ -95,6 +101,22 @@ export default class AutoSuggestField extends Component {
...
+    return { nodeType: props =>
+      (<Highlighter
+        indexHighlighted={indexHighlighted}
+        {...props}
```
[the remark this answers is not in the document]
nray621: On line 111?

packages/react-ui-core/src/Highlighter/__tests__/Highlighter-test.js
yasabere: These tests don't work with shallow

stories/core/List.js
```diff
@@ -34,16 +34,15 @@ export const ListWithPassedNodeTypes = (
...
     }}
   />
-
 )
 
 export const ListWithOwnItemComponent = (
   <List
     items={sampleListItems}
-    ListItem={props => (<SampleItemComponent {...props} />)}
+    listItem={props => (<SampleItemComponent {...props} />)}
```
tadjohnston: 👍

stories/theme/core/Highlighter.css
```diff
@@ -0,0 +1,8 @@
+.Highlighter {
+
+}
+
+.Highlighter_Match {
```
tadjohnston: `Match` is a modifier so it should be `.Highlighter-match`

stories/theme/core/AutoSuggestField.css
```diff
@@ -27,7 +27,15 @@
 }
 
 .AutoSuggestField .ListItem {
-  padding: 5px;
+  padding: 5px 0;
+}
+
+.AutoSuggestField .Highlighter_Match {
```
tadjohnston: `Match` is a modifier so it should be `.Highlighter-match`

packages/react-ui-core/src/Highlighter/Highlighter.js
```diff
@@ -0,0 +1,87 @@
...
+      indexHighlighted,
+      className,
+    } = this.props
+
+    if (!indexHighlighted || index === indexHighlighted) {
+      return (
+        <span
+          data-tid="highlighter-match"
+          className={classnames(
+            theme.Highlighter_Match,
```
tadjohnston: `Match` is a modifier so it should be `.Highlighter-match`

stories/theme/core/Highlighter.css
nray621: Gotcha. Changing now

packages/react-ui-core/src/Highlighter/Highlighter.js
```diff
@@ -0,0 +1,87 @@
...
+      pattern,
+      theme,
+      indexHighlighted,
+      className,
+    } = this.props
+
+    if (!indexHighlighted || index === indexHighlighted) {
+      return (
+        <span
+          data-tid="highlighter-match"
```
tadjohnston: Since this will be repeated, we may want to make the data-tid specific, but I'm not sure. Maybe:
```
data-tid=`highlighter-match-${index}`
```
I'm not sure if we _really_ need this, so please give thoughts on this.

nray621: @yasabere You a little more familiar with this code. What do you think?

packages/react-ui-core/src/Highlighter/Highlighter.js
yasabere: Done

packages/react-ui-core/src/index.js
```diff
@@ -48,6 +48,7 @@ export {
 
 export {
   List,
+  HighlightedList,
```
tadjohnston: This is not a component anymore is it?

nray621: Whoops. Sorry. I'll remove that

tadjohnston: Yeah, we need to be careful here. I think this would actually break things outside of this repo.

nray621: Gotcha. yeah, I'll be more careful with that going forward. That was my bad

PR #119, opened 2018-01-04
Add onPriceChange prop to PriceFilterCard

grenti:
[Card](https://rentpath.leankit.com/card/567588708)

* Added new onPriceChange prop to PriceFilterCard component
* Updated story examples with new prop func:
  * InlinePriceFilterCard: added example for single price
  * DesktopPriceFilterCard: added example for min/max price

### DIFF
--- a/packages/react-ui-ag/src/Filters/__tests__/PriceFilterCard-test.js
+++ b/packages/react-ui-ag/src/Filters/__tests__/PriceFilterCard-test.js
@@ -1,9 +1,11 @@
 import React from 'react'
 import { mount, shallow } from 'enzyme'
+import { RangeSlider as ThemedRangeSlider } from '@rentpath/react-ui-core'
 import FilterCard from '../FilterCard'
 import theme from './mocks/theme'
 import ThemedPriceFilterCard from '../PriceFilterCard'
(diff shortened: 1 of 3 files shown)


yasabere:
LGTM 👍

packages/react-ui-ag/src/Filters/__tests__/PriceFilterCard-test.js
```diff
@@ -35,4 +37,24 @@ describe('ag/Filters/PriceFilterCard', () => {
     expect(wrapper.find('[data-tid="foo"]')).toHaveLength(1)
     expect(wrapper.find('[data-tid="price-filter-card"]')).toHaveLength(0)
   })
+
+  it('calls onChange when <RangeSlider/> onChangeComplete event occurs', () => {
+    const onChangeFn = jest.fn()
+    const wrapper = mount(
```
tadjohnston: Do we need to `mount` here?

mneumark: He's searching for and calling a function a couple levels down, so in this case I think yes.

PR #120, opened 2018-01-04
Ratings filter card returns expected value

yasabere:
[card](https://rentpath.leankit.com/card/596093682)

-added checked so apply filter button returns value


danthamotheram:
👍

PR #121, opened 2018-01-06
Add Docker support

mneumark:
[Story](https://rentpath.leankit.com/card/595646843)

* create Dockerfile
* fix bug causing storybook static build to not work


Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+COPY --chown=node:node . $HOME/react-ui/
+RUN apt-get install -y nginx
+
+USER node
+RUN yarn config set registry https://registry.yarnpkg.com
+RUN yarn install
+
+ARG VERSION
+ARG COMMIT
+ARG BUILD_NUMBER
```
Christopher-Bui: Are you actually using these build args?

Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+
+RUN yarn run build-storybook
+
+USER root
+RUN cp -R storybook-static/* /var/www/html/
+
+EXPOSE 80
+
+CMD ["nginx", "-g", "daemon off;"]
```
Christopher-Bui: Could we switch back to the `node` user before running the container or do we need root?

Dockerfile
tadjohnston: We probably don't need these at all actually.

Dockerfile
mneumark: We need nginx to run as root. Just to be certain I just tried it locally and I get:
```

File: Dockerfile (hunk 1 above)
Comment (mneumark) [reply Dockerfile]: No, that was an area I thought we might talk about. Where does that stuff come in and do we need it for this type of repo?

**Comment by Christopher-Bui** [comment]
We'll also want to run lint and tests here.

File: Dockerfile (hunk 1 above)
Comment (mneumark) [reply Dockerfile]: Removed

File: script/dockerbuild [hunk 3]
```diff
@@ -0,0 +1,25 @@
...
+# Pass variables needed to generate BUILD-INFO
+docker build \
+      -f ./Dockerfile \
+      -t rentpath/react-ui:$VERSION \
+      --build-arg BRANCH=$BRANCH_NAME \
+      --build-arg BUILD_SHA=$GIT_COMMIT \
+      --build-arg BUILD_ORG_REPO=$REPO \
+      --build-arg BUILD_TARGET_URL=$BUILD_URL \
+      --build-arg BUILD_AUTH=$BUILD_AUTH \
+      --build-arg BUILD_NUMBER=$BUILD_NUMBER \
```
Christopher-Bui: You won't need these build args if you don't specify any in the dockerfile.

PR #122, opened 2018-01-08
Remove yarn.lock and lerna fixes

tadjohnston:
[Card](https://rentpath.leankit.com/card/597789445)

* Removed yarn.lock from every packages repo.  Added it to .gitignore.
This fixes an issue with lerna not updating yarn.lock which we really don't need anyway.  By removing this, it keeps us more in sync with what installing the package would do.
* Locked down package version to specific versions in all package.json
files
* Added publish script for lerna (adds exact version bump flag)
* Added css linter
* Fixed all css linting issues


packages/react-ui-ag/src/Filters/RadioGroupDropdown.js
```diff
@@ -93,10 +93,10 @@ export default class RadioGroupDropdown extends Component {
 }
 
 const RadioGroupDropdownAnchorTextComponent = props => (
-  <React.Fragment>
+  <div>
```
tadjohnston: Jest doesn't play nice when testing React.Fragments within the newest version ☹️ . Hopefully we can switch this back later.

lfender6445:
Im unclear on why we wouldn't want to lock down dependencies for running these packages in isolation?

tadjohnston:
What do you mean @lfender6445 ? I'm guessing you mean through a lock? When you install a package through rent-js for example, my understanding is that it will only use the package.json from the package. It would only lock down, for example `react-ui-core` when installing _within_ rent-js.  I don't think it ever looks at the `yarn.lock` file in the `react-ui-core` package.

lfender6445:
Yes via lock. 

I had missed that you updated the package.json's to lock down versions as well. 

My thoughts were 2 devs could potentially wind up with differnt versions of a dep like mapbox-gl but that is unlikely.

tadjohnston:
It's possible they could yes, but it would happen even with the yarn.lock files there. IMO doing what i did at least has a little less margin of that happening. Plus it makes it much easier to publish as lerna doesn't update the `yarn.lock` file anyway.

lfender6445:
Have you seen this tad? 

https://yarnpkg.com/blog/2017/08/02/introducing-workspaces/

lfender6445:
I'm fine with the approach as is, since we are being stricter on versioning inside the package-lock, so long as we are careful about making sure everyone who develops in this repo is resolving the same sets of dependencies. 

👍

tadjohnston:
@lfender6445 I remember them implementing workspaces, but haven't looked into them at all. Thanks for the article 👍 .  They look cool! I think I'll try to implement them next to see if we can possibly get rid of redundancy in our repos (like jest).

PR #125, opened 2018-01-09
Rename data-tid's of RadioFilterComponent

paudyalhari:
[Card](https://rentpath.leankit.com/card/596920021)

* Moved data-tid's from radioGroup input to label
* Updated data-tid's by appending with filter type

### DIFF
--- a/packages/react-ui-ag/src/Filters/RadioGroupFilterCard.js
+++ b/packages/react-ui-ag/src/Filters/RadioGroupFilterCard.js
@@ -2,6 +2,7 @@ import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 import classnames from 'classnames'
 import themed from 'react-themed'
+import omit from 'lodash/omit'
 import { RadioGroup } from '@rentpath/react-ui-core'
 import isEqual from 'lodash/isEqual'
 import autobind from 'autobind-decorator'
@@ -43,7 +44,17 @@ export default class RadioGroupFilterCard extends Component {
   }
 
   get fields() {
-    return this.props.fields || []
+    const { fields = [] } = this.props
+    return fields.map(field => {
+      const obj = field
+
+      if (typeof field.label === 'string') {
+        obj.label = { text: field.label, 'data-tid': `radiogroup-radiobutton-${field.label}` }
+      } else {
+        obj.label = Object.assign({ 'data-tid': `radiogroup-radiobutton-${field.label.text}` }, field.label)
+      }
+      return obj
+    })
   }
 
   currentlyCheckedValue(fields = this.fields) {
@@ -61,7 +72,7 @@ export default class RadioGroupFilterCard extends Component {
     const {
       className,
       theme,
-      fields,
+
       ...props
     } = this.props
 
@@ -78,9 +89,8 @@ export default class RadioGroupFilterCard extends Component {
       >
         <RadioGroup
           name="radio-group-filter-card-radio-group"
-          data-tid="radio-group-filter-card-radio-group"
           hideInputElement
-          fields={fields}
+          fields={this.fields}
           onChange={this.handleRadioGroupSelection}
         />
       </FilterCard>
--- a/packages/react-ui-core/src/Form/RadioGroup.js
+++ b/packages/react-ui-core/src/Form/RadioGroup.js
@@ -75,14 +75,15 @@ export default class RadioGroup extends Component {
 
   renderRadioButton(index, fieldProps) {
     const props = pick(this.props, ['name', 'hideInputElement', 'orientation'])
+
     return (
       <RadioButton
-        data-tid="radiogroup-radiobutton"
         key={`${this.id}-${index}`}
         checked={this.state.value === fieldProps.value}
         onChange={this.handleValueChange}
         {...omit(fieldProps, 'checked')}
         {...props}
+
       />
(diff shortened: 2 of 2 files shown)


tadjohnston:
I'd rather not add a `filterType`. This will set a precedent for adding `Desktkop` or `Mobile` for every component.  I think we should go over these changes.  We seem to be shoving things in rather than fixing the root of the issue.

packages/react-ui-core/src/Form/RadioGroup.js
```diff
@@ -75,14 +75,16 @@ export default class RadioGroup extends Component {
...
   renderRadioButton(index, fieldProps) {
     const props = pick(this.props, ['name', 'hideInputElement', 'orientation'])
+    const label = { 'data-tid': 'radiogroup-radiobutton', text: fieldProps.label }
     return (
       <RadioButton
-        data-tid="radiogroup-radiobutton"
         key={`${this.id}-${index}`}
         checked={this.state.value === fieldProps.value}
         onChange={this.handleValueChange}
+        label={label}
```
tadjohnston: You cannot change things like this in `react-ui-core`. This affects _everything_ that's using this component and you're now changing the structure of this label completely.  How does one add anything to label now? Won't this get overrided if I pass in a `label` prop?

packages/react-ui-ag/src/Filters/RadioGroupFilterCard.js
```diff
@@ -43,7 +45,18 @@ export default class RadioGroupFilterCard extends Component {
...
   get fields() {
-    return this.props.fields || []
+    const { fields = [], filterType = '' } = this.props
+    return fields.map(field => {
+      const obj = field
+
+      if (typeof field.label === 'string') {
+        obj.label = { text: field.label, 'data-tid': `radiogroup-radiobutton-${filterType}` }
+      } else {
+        obj.label = Object.assign({ 'data-tid': `radiogroup-radiobutton-${filterType}` }, field.label)
```
tadjohnston: What if the field is a function or a node? This doesn't work.

packages/react-ui-core/src/Form/RadioGroup.js
tadjohnston: Also, this is now *forcing* a label which isn't always needed.

packages/react-ui-ag/src/Filters/RadioGroupFilterCard.js
```diff
@@ -43,7 +44,17 @@ export default class RadioGroupFilterCard extends Component {
...
   get fields() {
-    return this.props.fields || []
+    const { fields = [] } = this.props
+    return fields.map(field => {
+      const obj = field
+
+      if (typeof field.label === 'string') {
+        obj.label = { text: field.label, 'data-tid': `radiogroup-radiobutton-${field.label}` }
+      } else {
+        obj.label = Object.assign({ 'data-tid': `radiogroup-radiobutton-${field.label.text}` }, field.label)
```
tadjohnston: You still cannot do this.  What if `label` is empty in the field or it's not an object?

tadjohnston:
Just put a `data-tid="label" on https://github.com/rentpath/react-ui/blob/master/packages/react-ui-core/src/Form/Label.js#L38 and call it good. 
```
    return (
      /* eslint-disable jsx-a11y/label-has-for */
      <label 
        data-tid="label"
        {...props} 
        className={classnames}
      >
        {text}{children}
      </label>
    )
```
The above allows label to be overwritten if one is passed in.  Also, by doing this it will only show up if label is used. In this PR, you're still _forcing_ label to be used even if it doesn't exist and that is incorrect.

packages/react-ui-core/src/Form/RadioGroup.js
```diff
@@ -75,14 +75,15 @@ export default class RadioGroup extends Component {
 
   renderRadioButton(index, fieldProps) {
     const props = pick(this.props, ['name', 'hideInputElement', 'orientation'])
+
     return (
       <RadioButton
-        data-tid="radiogroup-radiobutton"
```
tadjohnston: I'm not sure why you removed this.

packages/react-ui-ag/src/Filters/RadioGroupFilterCard.js
```diff
@@ -78,9 +89,8 @@ export default class RadioGroupFilterCard extends Component {
       >
         <RadioGroup
           name="radio-group-filter-card-radio-group"
-          data-tid="radio-group-filter-card-radio-group"
```
tadjohnston: 👍

packages/react-ui-core/src/Form/RadioGroup.js
paudyalhari: I think I accidentally removed it ?

paudyalhari: Ohh ok I removed this to add data-tid dynamically, but seems like I need to revert this back.

tadjohnston: You know, I think maybe removing this is fine as long as there is one on the RadioButton already.

paudyalhari: We dont have that in RadioButton Field, I think it would be better to have it here then.

PR #128, opened 2018-01-10
chore: add coverage and commitizen

tadjohnston:
[Card](https://rentpath.leankit.com/card/599519508)

changes the way we commit by adding structure

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ storybook-static
 .#*
 packages/**/yarn.lock
 packages/**/package-lock.json
+coverage
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
     "clean": "lerna run clean --yes",
     "build": "lerna run build",
     "test": "jest",
+    "test:coverage": "node node_modules/.bin/jest --coverage && cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js",
     "lint": "yarn run lint:js && yarn run lint:css",
     "lint:css": "stylelint \"**/*.css\"",
     "lint:css:fix": "stylelint \"**/*.css\" --fix",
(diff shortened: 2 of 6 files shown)


bethstevic:
Looks good, just a couple of questions. What will the commit structure be and is there a way to override it if we just want to commit a WIP that will be squashed before we push up?

tadjohnston:
@bethstevic great questions! You can still commit the way you were doing this before. Your initial commit should always involve using `git cz`.  We'll be following what commitizen does which is the following:

https://github.com/commitizen/cz-cli#using-the-command-line-tool

package.json
```diff
@@ -19,6 +20,12 @@
     "build-storybook": "build-storybook"
   },
   "jest": {
+    "collectCoverageFrom": [
+      "**/src/**/*.{js}",
```
lfender6445: this covers packages/foo/src?

.gitignore
```diff
@@ -8,3 +8,5 @@ storybook-static
 .#*
 packages/**/yarn.lock
 packages/**/package-lock.json
+.coveralls.yml
```
lfender6445: did you mean to commit and add to .gitignore?

lfender6445:
should any of these be added to .npmignore?

.gitignore
tadjohnston: Well.. probably not, but there's no reason to modify it after so we could just leave it as is.

package.json
tadjohnston: It collects from src based on the tests.  `**` is an N directories before `/src` so yes it is meant to cover everything in packages. We could add in each repo instead if that would be easier.

tadjohnston:
@lfender6445 good point, we should probably add `coverage` at least to `.npmignore`.

grenti:
This is awesome - LGTM 👍

lfender6445:
👍

colinrymer:
💯 love it!

PR #130, opened 2018-01-11
Create ListingCell for react-ui-ag

nray621:
[Card](https://rentpath.leankit.com/card/595782316)

- Paired with Omar Khan @okhanrt 
- Correct font and icons will be added in ag.js
- Styling in storybook matches [mocks](http://zpl.io/VkYWkJL)
- Carousel, FavoriteHeart, and Coupon are currently placeholders
- Renders one or two CTA buttons
- Passes click actions to CTA buttons, Favorite Heart, and whole ListingCell
- Clicks on CTA and Favorite Heart stopPropagation to prevent cell click from firing as well
- Takes custom RatingItem for RatingBar
- Works for single family listing


stories/ag/ListingCell.js
```diff
@@ -0,0 +1,55 @@
...
+      title: 'Cool Apartment',
+      bedroomText: '3 bedrooms',
+    }}
+    RatingItem={Star}
+    viewType="map"
+    ctaSection={[
+      {
+        children: '404-378-1428',
+        onClick: action('Phone CTA action'),
+        type: 'phone',
```
okhanrt: Is the type property of ctaSection ever being used? I don't see it being accessed anywhere else.

stories/ag/ListingCell.js
```diff
@@ -0,0 +1,55 @@
...
+      },
+      {
+        children: 'send an email',
+        onClick: action('contact CTA action'),
+        type: 'contact',
```
okhanrt: Ditto

stories/ag/ListingCell.js
nray621: yeah we use it for the `data-tid`

stories/ag/ListingCell.js
nray621: see above

packages/react-ui-ag/src/ListingCell/ListingCell.js
```diff
@@ -0,0 +1,179 @@
...
+    )
+  }
+
+  render() {
+    const {
+      theme,
+      viewType,
+      onCardClick,
+      listingDetails,
+      className,
```
yasabere: Would it make sense to have a favoriteItem?

nray621: Yeah, as long as we decide for sure that we aren't going to have a react-ui-ag favoriteHeart component and just use the react-ui-core toggleButton, then yeah. I will wait to make that change until that choice is finalized

stories/theme/ag/small/ListingCell.css
```diff
@@ -0,0 +1,118 @@
...
+
+.ListingCell_Favorite {
+  height: 16px;
+  width: 15px;
+  position: relative;
+  bottom: 135px;
+  left: 218px;
+}
+
+.ListingCell_Info_Bottom,
```
tadjohnston: You should never have 2 `_`s in a classname.  If you do, either the syntax is wrong, or your component can be broken down further.

stories/theme/ag/small/ListingCell.css
```diff
@@ -0,0 +1,118 @@
...
+.ListingCell_Info_Top {
+  display: flex;
+  justify-content: space-between;
+  align-items: flex-end;
+}
+
+.ListingCell_Info_Top {
```
tadjohnston: This is the same class as above, please combine.

PR #131, opened 2018-01-11
feat: adding core carousel

tadjohnston:
[Card](https://rentpath.leankit.com/card/594954095)
 
- added functionality for paging
- added navigation
- added photo carousel


packages/react-ui-core/src/Carousel/Carousel.js
```diff
@@ -0,0 +1,231 @@
...
+      className,
+      theme,
+      items,
+      navigation,
+      pagination,
+      showNav,
+      ...rest
+    } = this.props
+
+    console.log('nav: ', this.navigation)
```
nray621: leftover console.log

mneumark:
There are a couple of linting issues according to jenkins.

packages/react-ui-core/src/Carousel/PhotoCarousel.js
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  photo(path) {
+    const { server, sizes } = this.props
+    return `${server}${path}${sizes}`
+  }
+
+  lazyLoad() {
+    return (
+      <LazyLoad>
```
mneumark: I see you set up lazyLoad as either a bool or an object, but you never seem to use lazyLoad as an object anywhere. Oversight?

tadjohnston: I meant to add it in as props here, need to check if it's an object here first.

packages/react-ui-core/src/Carousel/Carousel.js
tadjohnston: Doh! good catch!

mneumark:
Looks good. I see tests are failing because there aren't any, but as far as I know that's getting handled in a different ticket by Beth/Anye. How were you planning to resolve that?

nray621:
LGTM 👍

PR #134, opened 2018-01-17
chore(script): publish script needs registry set

tadjohnston:
[Card](https://rentpath.leankit.com/card/602636449)
Fixing publish script. Breaks without registry when using yarn command


mneumark:
As far as I know, we have been using `https://registry.yarnpkg.com` elsewhere. Did we want to use that or the NPM one?

tadjohnston:
The npm registry has to be used to publish the package to https://www.npmjs.com/org/rentpath

nray621:
👍

PR #146, opened 2018-01-22
chore(travis): adding NODE_ENV for tests

tadjohnston:

Adding NODE_ENV for jest so travis works


tadjohnston:
Not needed.

PR #150, opened 2018-01-23
feat(Create Listing Cell component and ListingComponents): create ListingCell

mneumark:

[Story](https://rentpath.leankit.com/card/596951230)

* created generic listing cell that passes context down
* created text components that take context and uses that as text
* created wrapper to pass context as props for RatingBar and PhotoCarousel


yasabere:
LGTM 👍

PR #152, opened 2018-01-24
feat(adding feature): create mobile map listing

nray621:

make MobileMapListing from core ListingCell
create basic styling based on mocks
add event argument to toggleButton handle click to allow prevention of event bubbling
move banner story from example to small
Further styling and static elements (SVGs, fonts, carousel nav icons) will be added on implementation into ag.js
[card](https://rentpath.leankit.com/card/595782316)

### DIFF
--- a/stories/theme/ag/small/MobileMapListing.css
+++ b/stories/theme/ag/small/MobileMapListing.css
@@ -1,0 +1,131 @@
+.ListingCell {
+  border-radius: 5px;
+  border: 1px solid lightgrey;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+}
+
+.ListingCell .RatingBar_Items {
+  display: flex;
+  flex-direction: row;
+}
+
+.MobileMapListing {
+  width: 280px;
+  height: 200px;
+}
+
+.MobileMapListing_Top {
+  width: 100%;
+}
+
+.MobileMapListing_Bottom {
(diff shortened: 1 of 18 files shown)


bethstevic:
Looks good 👍

stories/theme/ag/small/MobileMapListing.css
```diff
@@ -0,0 +1,121 @@
...
+  font-family: ProximaNova;
+  color: #ffffff;
+  font-size: 11px;
+}
+
+.MobileMapListing .Button {
+  outline: none;
+}
+
+.MobileMapListing_FavoriteButton {
```
tadjohnston: You shouldn't need to do this. This would be better IMO:
```
.MobileMapListing .FavoriteButton {
```
that way you don't have to add a classname to everything. Things like FavoriteButton should already have one.

mneumark: We thought of that. We actually tried, but there is no `FavoriteButton` component, so to do that we'd have to modify the `themed` call at the top to something like:
```js
@themed(/^(MobileMapListing|FavoriteButton)/, { pure: true })
```

Would you rather us modify the call to `themed` or leave it as is?

yasabere:
LGTM 👍

bethstevic:
Looks good 👍

PR #153, opened 2018-01-25
fix(Switched to babel-preset-env to use async/await & regenerator plugins for async…

grenti:
… keyword causing issues on ag.j

[Card](https://rentpath.leankit.com/card/605433729)


nray621:
As long as we can verify that this doesn't break anything for anyone currently using react-ui, I'm good on this.

PR #154, opened 2018-01-25
fix(babel): updated babel preset to env from es2015.

grenti:

This resolved async Mapbox.init component call which wasn't getting transpiled to ES5
and causing
issues with ag.js

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "babel-plugin-transform-decorators-legacy": "1.3.4",
     "babel-plugin-transform-object-rest-spread": "6.26.0",
     "babel-plugin-transform-runtime": "6.23.0",
-    "babel-preset-es2015": "6.24.1",
+    "babel-preset-env": "1.6.1",
     "babel-preset-react": "6.24.1",
     "commitizen": "2.9.6",
     "coveralls": "3.0.0",
--- a/packages/react-ui-utils/package.json
+++ b/packages/react-ui-utils/package.json
@@ -19,7 +19,7 @@
     "babel-eslint": "8.1.2",
     "babel-jest": "22.0.4",
     "babel-plugin-transform-object-rest-spread": "6.26.0",
-    "babel-preset-es2015": "6.24.1",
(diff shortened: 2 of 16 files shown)


package.json
```diff
@@ -60,6 +60,7 @@
     "babel-plugin-transform-decorators-legacy": "1.3.4",
     "babel-plugin-transform-object-rest-spread": "6.26.0",
     "babel-plugin-transform-runtime": "6.23.0",
+    "babel-preset-env": "^1.6.1",
     "babel-preset-es2015": "6.24.1",
```
lfender6445: shouldn't we also be removing babel-preset-es2015

https://github.com/babel/babel/tree/master/packages/babel-preset-env

> Without any configuration options, @babel/preset-env behaves exactly the same as @babel/preset-latest (or @babel/preset-es2015, @babel/preset-es2016, and @babel/preset-es2017 together).

packages/react-ui-utils/package.json
```diff
@@ -19,7 +19,7 @@
     "babel-eslint": "8.1.2",
     "babel-jest": "22.0.4",
     "babel-plugin-transform-object-rest-spread": "6.26.0",
-    "babel-preset-es2015": "6.24.1",
+    "babel-preset-env": "^1.6.1",
```
lfender6445: i noticed all deps except this one are specifically locked in. it might be safer to follow same pattern

lfender6445:
would i expect to see yarn lock changes also committed? not sure how yarn handles dev deps. aside from comments I am 👍

package.json
grenti: @lfender6445 correct. That was an oversight on my part with these two packages. I'll update and push up. 👍

packages/react-ui-utils/package.json
grenti: @lfender6445 - I agree, I'll install the specific version. When the preset is updated with latest language additions we can upgrade it manually as we do with the other `yarn outdated` packages. 👍

PR #155, opened 2018-01-26
Yo dawg, I heard you like Babel...

colinrymer:
Update Babel (including plugins and presets) to the latest version and
remove all individual package `.babelrc` files as they aren't needed. By
default, Babel will walk up the directory tree until it finds a `.babelrc`
file, which means that we can leverage just one `.babelrc` file in the root
of our monorepo.

Node module resolution works in a similar manner; if node is unable to find
a package, it walks up the directory tree looking for `node_module` directories
that have the missing packages. As such, we can install all the dev deps
in our monorepo's root `package.json` and remove all the dev deps from the
individual packages' `package.json` files. This makes managing dev deps much
easier, as they now live in one file and we can use tooling like Greenkeeper
to keep them up to date.


nray621:
nice 👍

colinrymer:
Actually going to put this one on hold for a bit; looks like storybook doesn't yet support Babel 7. 😐

.babelrc
```diff
@@ -1,12 +1,11 @@
 {
   "presets": [
-    "es2015",
-    "react"
+    "@babel/preset-env",
+    "@babel/preset-react",
+    "@babel/preset-stage-0"
```
justinrex: screen shot 2018-06-07 at 10 50 31 am

PR #156, opened 2018-01-26
DatePicker 

pkanuganti:
[Story](https://rentpath.leankit.com/card/606922844)

### DIFF
--- a/packages/react-ui-core/src/DateInput/DateInput.js
+++ b/packages/react-ui-core/src/DateInput/DateInput.js
@@ -1,0 +1,24 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+import DatePicker from 'react-datepicker'
+
+export default class DateInput extends Component {
+
+ static propTypes = {
+   selected: PropTypes.object,
+   onChange: PropTypes.func,
+   maxDate: PropTypes.object,
+   minDate: PropTypes.object,
+   placeholderText: PropTypes.string,
+ }
+
+ static defaultProps = {
+   placeholderText: 'Date',
+ }
+
+ render() {
+   return (
+     <DatePicker {...this.props} />
+   )
+ }
+}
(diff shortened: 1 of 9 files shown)


packages/react-ui-core/src/DateInput/DateInput.js
```diff
@@ -0,0 +1,24 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+import DatePicker from 'react-datepicker'
+
+export default class DateInput extends Component {
+
```
bethstevic: Do we want to go ahead and set up the `themed` like the new components we have been writing? Something like `@themed(['DateInput'])`

bethstevic:
Looks good 👍

packages/react-ui-core/src/DateInput/DateInput.js
nray621: +1

pkanuganti: Datepicker is an external lib. So all the css class are global . so we don't need to have setup like @themed(['DateInput'])

packages/react-ui-core/src/DateInput/DateInput.js
```diff
@@ -0,0 +1,24 @@
+import React, { Component } from 'react'
```
nray621: Isn't this just an empty wrapper that passes props through to the `DatePicker`?
Would something closer to what you wrote for the story's component maybe be better? idk, just throwing that out there

pkanuganti: Its just wrapper to pass props to DatePicker. In story book, we have default behaviour for Datepicker.

packages/react-ui-core/src/DateInput/DateInput.js
nray621: You should still have it imo

nray621:
ok 👍

PR #159, opened 2018-01-29
fix(Drawer update): component now spreads props to toggleButton compo…

yasabere:
…nent


bethstevic:
Looks good 👍 , can you fix your commit message though?

PR #160, opened 2018-01-29
Ml yeoman

WinstonMain:
[story](https://rentpath.leankit.com/card/595844382)

POC for using Yeoman with React-UI


colinrymer:
@WinstonMain why was this closed?

PR #163, opened 2018-01-31
fix(Hybrid Map Listing): fixed ratings in ListingCell for Listing Car…

grenti:
…ousel for the Hybrid Map.

[Card](https://rentpath.leankit.com/card/607050974)

### DIFF
--- a/stories/core/ListingCell.js
+++ b/stories/core/ListingCell.js
@@ -15,10 +15,14 @@ const baseListing = {
   rating: {
     score: 4,
     label: '20',
-    fillColor: 'yellow',
-    backgroundFillColor: '#ffffff',
     uniqueId: '123',
   },
+  phone: '555-867-5309',
+}
+
+const ratingProps = {
+  fillColor: 'yellow',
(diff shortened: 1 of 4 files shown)


stories/core/ListingCell.js
```diff
@@ -32,7 +36,7 @@ export const DefaultListingCell = (
       <div>
         <ListingComponents.Bathroom />
         <h1>hello</h1>
-        <ListingComponents.Ratings />
+        <ListingComponents.Ratings {...ratingProps} />
```
nray621: Why are we changing this to pass props directly to the ratings listing component?
Isn't the point that they just get their props through the `ListingCell`?

mneumark: The idea is to give you the ability to both. So settings like fill color and such can be passed as props. And then the rating score and label will go in via context through `ListingCell`.

nray621: ok cool 👍

nray621:
👍  LGTM

bethstevic:
Looks good 👍

PR #166, opened 2018-02-01
chore(all): leverage shared devdeps and scripts

colinrymer:
@rentpath/react-ui-rentals, @rentpath/react-ui-tracking, @rentpath/react-ui-utils

This commit removes the devDependencies from each package and merges them into
the devDependencies of the root package. Because of how Node handles module
resolution, we can leverage the shared node_modules directory of a parent dir
when looking for modules during code execution (in development). Any actual
dependencies still remain in their respective package.json files.

Additionally, this commit removes the scripts from each package and updates the
root package.json's scripts to leverage `lerna exec` instead of `lerna run`
which will execute the given command in each package, rather than looking for a
script of that name in each package as `lerna run` does. Because of this change,
we can further simplify each package by removing the `.babelrc` and the
jest configuration/setup from each package.

One of the biggest benefits to this code reorganization is that we can now
easily manage keeping our development dependencies up to date using Greenkeeper,
etc. as those services only work with a single package.json at the root of the
repo.

Furthermore, any changes to our test config, etc. will automatically be shared
across all our packages.

A common concern with these changes is that the tests are no longer isolated to
a single package, but in practice, the tests run quickly enough that this should
not be an issue. On my development machine, I see full runs of the test suite
regularly finish in under 10 seconds. Utilizing the --watch flag, etc. to limit
the tests to just the files changed will result in even faster times.


lfender6445:
Checking out the branch fresh, and running `yarn`

```
SyntaxError: src/Mapbox/Mapbox.js: Unexpected token (98:10)
   96 |
   97 |   async init() {
>  98 |     await import(/* webpackChunkName: "mapbox" */ 'mapbox-gl/dist/mapbox-gl').then(mapbox => {
      |           ^
   99 |       this.MapboxGL = mapbox
  100 |       this.map = this.setupMapbox()
  101 |
lerna ERR! exec Errored while executing 'babel --ignore __tests__ -d lib src' in '@rentpath/react-ui-core'
lerna ERR! execute callback with error
lerna ERR! Error: Command failed: babel --ignore __tests__ -d lib src
lerna ERR!     at Promise.all.then.arr (/Users/lfender/source/react-ui/node_modules/execa/index.js:236:11)
lerna ERR!     at <anonymous>
{ Error: Command failed: babel --ignore __tests__ -d lib src
    at Promise.all.then.arr (/Users/lfender/source/react-ui/node_modules/execa/index.js:236:11)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7)
```

colinrymer:
@lfender6445 yeah, seeing that issue in Travis as well. I know the cause of the issue; not sure how to fix it with the version of babel we're using, but we can't upgrade babel versions because storybook doesn't support babel 7.

All that to say: word - I'm working on a fix.

ninomichaels:
it doesn't  look like `build:watch` works in this configuration yet. when we initially setup lerna, devDeps were shared, and this is the reason why it was split up.

They've since added `lerna run --parallel`, so we can do `lerna run --parallel build:watch`, but this still requires that command to be in their respective `package.json` files. not sure how it feels about the shared deps though.

tadjohnston:
Fixed the await import issue.  This is due to the way webpack wants the dynamic import (`require.ensure()`) vs how node uses it (straight forward). This is the reason we have a test env for our babelrc in react-ui-core.

test mode needs:
```
        "dynamic-import-node"
```

build mode needs:
```
        "dynamic-import-webpack"
```

PR #169, opened 2018-02-08
Set up Prioritize Card Click for CTA buttons on Listing Carousel Drawer

paudyalhari:
[Card](https://rentpath.leankit.com/card/613128248)

* Added login to handle clicks
* Added Unit Test

### DIFF
--- a/packages/react-ui-ag/src/Listings/MobileMapListing.js
+++ b/packages/react-ui-ag/src/Listings/MobileMapListing.js
@@ -48,10 +48,14 @@ export default class MobileMapListing extends PureComponent {
 
   @autobind
   handleButtonClick(onClick) {
-    const { prioritizeCardClick } = this.props
-
     return event => {
-      if (!prioritizeCardClick && onClick) onClick(this.props.listing)
+      const { prioritizeCardClick } = this.props
+
+      if (!prioritizeCardClick && onClick) {
+        onClick(this.props.listing)
+      } else if (prioritizeCardClick && this.props.onClick) {
+        this.props.onClick(this.props.index)
+      }
 
       const shouldStopPropagation =
         !prioritizeCardClick && event && event.stopPropagation
--- a/packages/react-ui-ag/src/Listings/SingleFamilyMobileMapListing.js
+++ b/packages/react-ui-ag/src/Listings/SingleFamilyMobileMapListing.js
@@ -42,10 +42,14 @@ export default class SingleFamilyMobileMapListing extends PureComponent {
 
   @autobind
   handleButtonClick(onClick) {
-    const { prioritizeCardClick } = this.props
-
     return event => {
-      if (!prioritizeCardClick && onClick) onClick(this.props.listing)
+      const { prioritizeCardClick } = this.props
+
+      if (!prioritizeCardClick && onClick) {
+        onClick(this.props.listing)
+      } else if (prioritizeCardClick && this.props.onClick) {
+        this.props.onClick(this.props.index)
+      }
 
       const shouldStopPropagation =
         !prioritizeCardClick && event && event.stopPropagation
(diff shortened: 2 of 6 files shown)


packages/react-ui-ag/src/Listings/MobileMapListing.js
```diff
@@ -48,10 +48,14 @@ export default class MobileMapListing extends PureComponent {
...
   handleButtonClick(onClick) {
-    const { prioritizeCardClick } = this.props
-
     return event => {
-      if (!prioritizeCardClick && onClick) onClick(this.props.listing)
+      const { prioritizeCardClick } = this.props
+
+      if (!prioritizeCardClick && onClick) {
+        onClick(this.props.listing)
+      } else if (prioritizeCardClick) {
```
mneumark: I think you need another check in this else if to check if the onClick prop exists

packages/react-ui-ag/src/Listings/SingleFamilyMobileMapListing.js
```diff
@@ -42,10 +42,14 @@ export default class SingleFamilyMobileMapListing extends PureComponent {
...
   handleButtonClick(onClick) {
-    const { prioritizeCardClick } = this.props
-
     return event => {
-      if (!prioritizeCardClick && onClick) onClick(this.props.listing)
+      const { prioritizeCardClick } = this.props
+
+      if (!prioritizeCardClick && onClick) {
+        onClick(this.props.listing)
+      } else if (prioritizeCardClick) {
```
mneumark: I think you need another check in this else if to check if the onClick prop exists

packages/react-ui-ag/src/Listings/__tests__/MobileMapListing-test.js
```diff
@@ -92,6 +94,19 @@ describe('ag/Listing/MobileMapListing', () => {
       wrapper.find('[data-tid="cta-button"]').at(1).simulate('click')
       expect(ctaClick.mock.calls).toHaveLength(2)
     })
+
+    it('fire cta buttons on click action when prioritizeCardClick is true', () => {
```
mneumark: add an `s` to the end of `fire`

packages/react-ui-ag/src/Listings/__tests__/SingleFamilyMobileMapListing-test.js
```diff
@@ -60,6 +60,20 @@ describe('ag/Listing/SingleFamilyMobileMapListing', () => {
     expect(ctaClick).toHaveBeenCalled()
   })
 
+  it('fire cta buttons on click action when prioritizeCardClick is true', () => {
```
mneumark: add an `s` to the end of `fire`

packages/react-ui-ag/src/Listings/__tests__/SingleFamilyMobileMapListing-test.js
```diff
@@ -60,6 +60,20 @@ describe('ag/Listing/SingleFamilyMobileMapListing', () => {
     expect(ctaClick).toHaveBeenCalled()
   })
 
+  it('fires cta buttons on click action when prioritizeCardClick is true', () => {
```
bethstevic: Do you think we could add some tests to make sure it doesn't fire the click actions when prioritizeCardClick is false? Since that is the issue we were having before

bethstevic:
Looks good 👍

packages/react-ui-ag/src/Listings/__tests__/SingleFamilyMobileMapListing-test.js
mneumark: @bethstevic Just talked to @paudyalhari . Seems we do seem to have those tests. Just they are in ListingCell, not here.

PR #175, opened 2018-02-10
fix(Hybrid Map): fix how onSlide is handled by ListingCarousel

mneumark:

[Story](https://rentpath.leankit.com/card/615319754)

* moved call to onSlide prop to updateSelectedIndex
* removed onSlide in render to avoid it overriding stuff in Carousel


nray621:
LGTM 👍

PR #176, opened 2018-02-10
refactor(react-ui-core): change drawer's animation to use translateY

nray621:

### DIFF
--- a/packages/react-ui-core/src/Drawer/Drawer.js
+++ b/packages/react-ui-core/src/Drawer/Drawer.js
@@ -1,5 +1,4 @@
 import React, { Component } from 'react'
-import { Motion, spring } from 'react-motion'
 import themed from 'react-themed'
 import PropTypes from 'prop-types'
 import classnames from 'classnames'
@@ -27,7 +26,7 @@ export default class Drawer extends Component {
 
   static defaultProps = {
     theme: {},
-    height: 220,
+    height: 205,
     movementDamping: null,
     movementStiffness: null,
     visible: true,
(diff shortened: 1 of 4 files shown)


packages/react-ui-core/src/Drawer/Drawer.js
```diff
@@ -72,42 +71,33 @@ export default class Drawer extends Component {
...
-                    height: `${height - x}px`,
-                  }}
-                >
-                  {children}
-                </div>}
-              </div>
-            )
-          }
-        </Motion>
+        <div>
```
mneumark: What is the reason for this div?

packages/react-ui-core/src/Drawer/Drawer.js
```diff
@@ -72,42 +71,33 @@ export default class Drawer extends Component {
 
     return (
       <div
-        className={classnames(theme.Drawer, className)}
+        className={classnames(
+          theme.Drawer,
+          theme[this.state.visible ? 'Drawer-on' : 'Drawer-off'],
```
mneumark: You seem to use this.state.visible a lot. Maybe pull it out above?

nray621: yup. good call

nray621:
I don't know that it is react motion being unusable, but it seemed to be having trouble animating heavier things like the listing carousel. To be honest I haven't really looked into why exactly this is smoother, but I just changed it to animate the same way rent-js animates their listing carousel.

PR #178, opened 2018-02-11
fix(react-ui-ag): render an empty div instead of ratings if no rating…

nray621:
… present


danthamotheram:
LGTM 👍

yasabere:
LGTM 👍

PR #182, opened 2018-02-12
fix(Hybrid Map): reverted href attribute changes

grenti:

[Card](https://rentpath.leankit.com/card/613135571)
- Reverted href changes because action got
triggered on inactive card in carousel


yasabere:
LGTM :thumpsup:

bethstevic:
Looks good 👍

PR #183, opened 2018-02-12
Edit photo carousel to accept data tags

bethstevic:
Allow an object to be passed to item in the photo carousel props


nray621:
LGTM 👍

PR #190, opened 2018-02-21
feat(Desktop Hybrid Map): create DesktopMapPinListing for Desktop Hybid Map

mneumark:

[Story](https://rentpath.leankit.com/card/619763011)

* create `DesktopMapPinListing`
* update ListingComponents/ListingCell with `Neighborhood` and `Photo` components

### DIFF
--- a/packages/react-ui-ag/src/Listings/DesktopMapPinListing.js
+++ b/packages/react-ui-ag/src/Listings/DesktopMapPinListing.js
@@ -1,0 +1,153 @@
+import React, { PureComponent } from 'react'
+import PropTypes from 'prop-types'
+import themed from 'react-themed'
+import classnames from 'classnames'
+import autobind from 'autobind-decorator'
+import get from 'lodash/get'
+import { Button, ToggleButton, ListingComponents, ListingCell } from '@rentpath/react-ui-core'
+
+const buttonPropTypes = PropTypes.shape({
+  children: PropTypes.node,
+  onClick: PropTypes.func,
+  className: PropTypes.string,
+  valueLocation: PropTypes.string,
+})
+
+@themed(/^DesktopMapPinListing/, { pure: true })
+export default class DesktopMapPinListing extends PureComponent {
+  static propTypes = {
+    index: PropTypes.number,
+    listing: PropTypes.object,
+    theme: PropTypes.object,
+    className: PropTypes.string,
+    onClick: PropTypes.func,
+    photo: PropTypes.object,
+    propertyName: PropTypes.object,
+    ctaButtons: PropTypes.arrayOf(buttonPropTypes),
+    favoriteButton: buttonPropTypes,
+  }
+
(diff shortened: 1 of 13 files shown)


packages/react-ui-ag/src/Listings/DesktopMapPinListing.js
```diff
@@ -0,0 +1,155 @@
...
+        listing={listing}
+        onClick={this.handleCardClick}
+        className={classnames(
+          className,
+          theme.DesktopMapPinListing,
+        )}
+        {...props}
+      >
+        <ListingComponents.Photo {...photo} />
+        <div className={theme.DesktopMapPinListing_Details}>
```
tadjohnston: What do you think about trying to get rid of some of the divs?  I think we can just add css to replace them.  It would help get rid of some of the extra dom nodes, especially when we have 30+ ( or even 100) listings.

packages/react-ui-ag/src/Listings/DesktopMapPinListing.js
```diff
@@ -0,0 +1,155 @@
...
+
+  renderFavoriteButton() {
+    const { theme, favoriteButton, listing } = this.props
+    const { className } = favoriteButton
+    return (
+      <ToggleButton
+        {...favoriteButton}
+        value={listing.isFavorited}
+        className={classnames(
+          theme.DesktopMapPinListing_FavoriteButton,
```
tadjohnston: Do we need things like `theme.DesktopMapPinListing_FavoriteButton` to be passed in? Why not just have one classname for the DesktopMapPinListing and over qualify them for things like FavoriteButton?

tadjohnston: I guess either way it's fine, but maybe we should have a FavoriteButton component.  That would make sense to me especially if we keep using this same code in different places.

mneumark: I guess we didn't because there really isn't any logic in a favorite button that isn't in a ToggleButton. The only reason would be for styling. If you think that would be a good enough reason that's fine.

packages/react-ui-ag/src/Listings/DesktopMapPinListing.js
mneumark: If you'd like to work with me on eliminating divs I'm happy to work with you on that.

tadjohnston: It would be preferable as we'd like to keep the number down.

packages/react-ui-ag/src/Listings/DesktopMapPinListing.js
tadjohnston: Whatever you think I'm 👍 .

paudyalhari:
👍

PR #191, opened 2018-02-21
577743353 Add allowUnselect flag to RadioGroup

barelyfitz:
LeanKit: [rent-js desktop more filters](https://rentpath.leankit.com/card/577743353)

To handle a use case when we have a group of radio buttons, but user also needs ability to unselect all the buttons:
- Added `allowUnselect` prop to the RadioGroup component
- Added `onUnselect` callback, because setting state to unselect the radio group cannot accurately trigger the `onChange` event. So if you are using allowUnselect it is on you to listen for both events if necessary.
- Added demo `Unselect RadioGroup` to the storybook
- Add unit tests for RadioGroup component
  - Verify button can be clicked to unselect when using allowUnselect
  - Verify button cannot be clicked to unselect when not using allowUnselect
  - Verify onUnselect / onChange event handlers are called as appropritate

### DIFF
--- a/packages/react-ui-core/src/Form/RadioGroup.js
+++ b/packages/react-ui-core/src/Form/RadioGroup.js
@@ -25,12 +25,15 @@ export default class RadioGroup extends Component {
       checked: PropTypes.bool,
       value: PropTypes.string,
     })),
+    allowUnselect: PropTypes.bool,
     onChange: PropTypes.func,
+    onUnselect: PropTypes.func,
   }
 
   static defaultProps = {
     theme: {},
     fields: [],
+    allowUnselect: false,
   }
 
   constructor(props) {
@@ -73,6 +76,16 @@ export default class RadioGroup extends Component {
     }
   }
 
+  handleClick = event => {
+    // Check if value was already selected, should we unselect it?
+    if (this.props.allowUnselect && this.state.value === event.target.value) {
+      this.setState({ value: null })
+      if (this.props.onUnselect) {
+        this.props.onUnselect(event)
+      }
+    }
+  }
+
   renderRadioButton(index, fieldProps) {
     const props = pick(this.props, ['name', 'hideInputElement', 'orientation'])
 
@@ -82,6 +95,7 @@ export default class RadioGroup extends Component {
         key={`${this.id}-${index}`}
         checked={this.state.value === fieldProps.value}
         onChange={this.handleValueChange}
+        onClick={this.handleClick}
         {...omit(fieldProps, 'checked')}
         {...props}
       />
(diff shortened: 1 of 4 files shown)


packages/react-ui-core/src/Form/RadioGroup.js
```diff
@@ -73,6 +77,19 @@ export default class RadioGroup extends Component {
...
 
+  handleClick = event => {
+    // Check if value was already selected, should we unselect it?
+    if (this.props.allowUnselect && this.state.value === event.target.value) {
+      this.setState({ value: null })
+      if (this.props.onUnselect) {
+        this.props.onUnselect(event)
+      }
+    }
+    if (this.props.onClick) {
```
tadjohnston: This is causing issues.  If something changes, this gets fired twice if I add an onClick prop
```
export const UnselectRadioGroup = (
  <RadioGroup
    name={'radio-group-4'}
    orientation="vertical"
    hideInputElement
    fields={[
      { label: 'Uno', checked: true, value: 'Uno' },
      { label: 'Dos', checked: false, value: 'Dos' },
    ]}
    onChange={event => action('change')(event.target.value)}
    allowUnselect
    onClick={event => action('select')(event.target.value)}
    onUnselect={event => action('unselect')(event.target.value)}
  />
)
```

screen shot 2018-02-22 at 8 04 11 am

I don't think `onClick` needs to be added to this component.

barelyfitz: @tadjohnston Thanks, I added the `onClick` prop because I thought I was "consuming" the onclick event from the RadioButton and needed to pass it through in case someone wanted an `onClick` handler for the RadioGroup, but that doesn't seem to be the case. I will remove.

If you think everything else is okay, I'll move forward with writing tests.

tadjohnston: Yeah, I figured that's why you did it. I had to 2nd guess myself as well and look at it a couple times. 😄

packages/react-ui-core/src/Form/__tests__/RadioGroup-test.js
```diff
@@ -156,4 +156,52 @@ describe('Form/RadioGroup', () => {
...
+      onUnselect: spyOnUnselect,
+      allowUnselect: true,
+    }
+    const wrapper = mount(<RadioGroup {...props} />)
+    expect(wrapper.state('value')).toEqual('Two')
+    wrapper.find('input[value="Two"]').simulate('click')
+    expect(wrapper.state('value')).toEqual(null)
+    expect(wrapper.find('input[value="Two"]').props()).not.toHaveProperty('checked', true)
+    expect(wrapper.find('input[value="One"]').props()).not.toHaveProperty('checked', true)
+    expect(spyOnChange).not.toBeCalled()
```
tadjohnston: 👍 nice check here.

nray621:
Nice addition 👍 LGTM

PR #192, opened 2018-02-22
feat(Desktop Hybrid Map): create DropdownMenu component

okhanrt:

[Card](https://rentpath.leankit.com/card/624767748)

* Create DropdownMenu component
* Add class for Dropdown_Body

### DIFF
--- a/packages/react-ui-core/src/Menu/DropdownMenu.js
+++ b/packages/react-ui-core/src/Menu/DropdownMenu.js
@@ -1,0 +1,78 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+import autobind from 'autobind-decorator'
+import themed from 'react-themed'
+import classnames from 'classnames'
+import { Dropdown } from '../Dropdown'
+import MenuWrapper from './MenuWrapper'
+
+@themed(/^DropdownMenu/)
+
+export default class DropdownMenu extends Component {
+  static propTypes = {
+    theme: PropTypes.object,
+    options: PropTypes.array,
+    onItemSelect: PropTypes.func,
+    className: PropTypes.string,
+    selectedIndex: PropTypes.number,
+    createAnchorText: PropTypes.func,
+  }
+
+  static defaultProps = {
+    theme: {},
+    options: [],
+    selectedIndex: 0,
+    onItemSelect: () => {},
+  }
+
+  constructor(props) {
+    super(props)
+    this.state = { selectedIndex: this.props.selectedIndex || 0 }
+  }
+
+  get selectedItem() {
+    return this.props.options[this.state.selectedIndex] || ''
+  }
+
+  @autobind
+  itemSelectionHandler(item, index) {
+    const { onItemSelect } = this.props
+
+    if (onItemSelect) onItemSelect(item, index)
+
+    this.setState({ selectedIndex: index })
+  }
+
+  renderAnchorFieldText() {
+    const { createAnchorText } = this.props
+
+    if (createAnchorText) return createAnchorText(this.selectedItem)
+
+    return this.selectedItem
+  }
+
+  render() {
+    const {
+      theme,
+      className,
+      options,
+      selectedIndex,
+      onItemSelect,
+      createAnchorText,
+      ...props
+    } = this.props
+
+    return (
+      <Dropdown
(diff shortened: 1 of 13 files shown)


packages/react-ui-core/src/Menu/MenuWrapper.js
```diff
@@ -0,0 +1,35 @@
+import React, { PureComponent } from 'react'
```
nray621: Maybe I'm missing something, but it seems like this component is unnecessary. The only addition here is `onSelect`. However, in `DropDownMenu` lines 72-73, you aren't passing a `onSelect` prop, so it is never going to be fired anyway.

packages/react-ui-core/src/Menu/DropdownMenu.js
```diff
@@ -0,0 +1,78 @@
...
+      ...props
+    } = this.props
+
+    return (
+      <Dropdown
+        className={classnames(className, theme.DropdownMenu)}
+        anchorField={{ children: this.renderAnchorFieldText() }}
+        {...props}
+      >
+        <MenuWrapper
```
nray621: Couldn't this just be:
```
        <Menu
          options={options}
          onItemSelect={this.itemSelectionHandler}
        />
```

packages/react-ui-core/src/Menu/__tests__/MenuWrapper-test.js
```diff
@@ -0,0 +1,19 @@
...
+import Menu from '../Menu'
+
+describe('MenuWrapper', () => {
+  it('calls the onItemSelect function when Menu onItemSelect is called', () => {
+    const onItemSelect = jest.fn()
+    const wrapper = shallow(<MenuWrapper onItemSelect={onItemSelect} />)
+    wrapper.find(Menu).props().onItemSelect()
+    expect(onItemSelect).toHaveBeenCalled()
+  })
+  it('calls the onSelect function when when Menu onItemSelect is called', () => {
```
nray621: nitpick: "when when"

packages/react-ui-core/src/Menu/MenuWrapper.js
mneumark: So, its not obvious, but take a look at the `Dropdown` component. It actually clones its children and passes an `onSelect` into it.

packages/react-ui-core/src/Menu/DropdownMenu.js
mneumark: See the comment in the `MenuWrapper` file.

mneumark:
@yasabere Yes it can, which is exactly what we did. But its a common enough pattern that having it pre-built makes it easier for re-use.

nray621:
LGTM 👍

bethstevic:
Looks good 👍

packages/react-ui-core/src/Menu/Menu.js
```diff
@@ -116,7 +132,7 @@ export default class Menu extends PureComponent {
           className,
           theme.Menu
         )}
-        items={options}
+        items={this.optionLabels}
```
tadjohnston: Seems a little specific to call this`optionLabels` doesn't it?  How about we just combine the options getter and make one getter?  We're only seem to be referencing `this.options` in one place.

packages/react-ui-core/src/Menu/DropdownMenu.js
```diff
@@ -31,7 +31,11 @@ export default class DropdownMenu extends Component {
   }
 
   get selectedItem() {
-    return this.props.options[this.state.selectedIndex] || ''
+    return this.itemLabel(this.props.options[this.state.selectedIndex]) || ''
```
tadjohnston: Still weird to me why we need to do all this rather than just pass back the object, but :shrug:

mneumark: Perhaps the name of the function could be better. The selectedItem is only used for the dropdown anchor text.

packages/react-ui-core/src/Menu/Menu.js
mneumark: Actually there are a couple other places that use options that should use the options getter, so rather than combine these, I'm going to actually extend the use of the options getter.

PR #195, opened 2018-02-27
fix(ListingCell, Drawer, MobileMapListing, HybridMobileMap): react-ui…

yasabere:
… optimizations

Gave MobileMapListing and SingleFamilyListing a 'shouldComponentUpdate' that uses lodash isEqual
against listing
Made Drawer pure
Made ListingCarousel pure

### DIFF
--- a/packages/react-ui-ag/src/Carousels/ListingCarousel.js
+++ b/packages/react-ui-ag/src/Carousels/ListingCarousel.js
@@ -1,4 +1,4 @@
-import React, { Component } from 'react'
+import React, { PureComponent } from 'react'
 import PropTypes from 'prop-types'
 import classnames from 'classnames'
 import { themed } from 'react-themed'
@@ -9,8 +9,10 @@ import {
   SingleFamilyMobileMapListing,
 } from '../Listings'
 
-@themed(/^ListingCarousel/)
-export default class ListingCarousel extends Component {
+@themed(/^ListingCarousel/,
+  { pure: true }
+)
+export default class ListingCarousel extends PureComponent {
   static propTypes = {
     listings: PropTypes.arrayOf(PropTypes.object),
     className: PropTypes.string,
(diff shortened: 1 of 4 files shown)


packages/react-ui-ag/src/Carousels/ListingCarousel.js
```diff
@@ -10,7 +10,7 @@ import {
 } from '../Listings'
 
 @themed(/^ListingCarousel/)
```
bethstevic: Make sure you add a `{ pure: true }` to the themed arguments

packages/react-ui-ag/src/Listings/MobileMapListing.js
```diff
@@ -15,7 +16,7 @@ const buttonPropTypes = PropTypes.shape({
 })
 
 @themed(/^MobileMapListing/, { pure: true })
-export default class MobileMapListing extends PureComponent {
+export default class MobileMapListing extends Component {
```
bethstevic: Also take this `{ pure: true }` off

packages/react-ui-ag/src/Carousels/ListingCarousel.js
bethstevic: Also, do you still need this check: `componentWillReceiveProps(nextProps) {
    if (nextProps.selectedIndex !== this.props.selectedIndex) {
      this.updateSelectedIndex(nextProps.selectedIndex)
    }
  }`
Since this is now a pure component?

packages/react-ui-ag/src/Listings/SingleFamilyMobileMapListing.js
```diff
@@ -14,7 +15,7 @@ const buttonPropTypes = PropTypes.shape({
 
 @themed(/^MobileMapListing/, { pure: true })
 
-export default class SingleFamilyMobileMapListing extends PureComponent {
+export default class SingleFamilyMobileMapListing extends Component {
```
bethstevic: Same here with line 16

packages/react-ui-ag/src/Carousels/ListingCarousel.js
mneumark: Yes. The only difference between `PureComponent` and `Component` is that `shouldComponentUpdate` is done for you and is a shallow comparison of all props. So basically by changing this to a `PureComponent` he's ensuring that props are compared shallowly rather than not at all.

bethstevic:
Looks 👍

PR #198, opened 2018-02-28
feat(Hybrid Map): Enable Lazyloading in MobileMapListing 

grenti:

Enabling react-image-gallery -lazy to render only 3 items for PhotoCarousel
Enabling react-lazyload to use placeholders for photos not needing rendering

### DIFF
--- a/packages/react-ui-ag/src/Carousels/__tests__/ListingCarousel-test.js
+++ b/packages/react-ui-ag/src/Carousels/__tests__/ListingCarousel-test.js
@@ -176,13 +176,7 @@ const listingProps = {
     dimensions: '280-120',
     disableSwipe: true,
   },
-  /* eslint-disable max-len */
-  // TODO
-  // gaa: This needs to be false because jsdom is providing a mocked DOM and react-lazyload
-  // manipulates the DOM so it crashes when attempting to create placeholders
-  // I'll need to look into this further later to see if we can fix this.
   lazyLoad: false,
-  /* eslint-enable max-len */
 }
 
 describe('ListingCarousel', () => {
--- a/packages/react-ui-ag/src/Carousels/__tests__/ListingCarousel-test.js
+++ b/packages/react-ui-ag/src/Carousels/__tests__/ListingCarousel-test.js
@@ -176,6 +176,13 @@ const listingProps = {
     dimensions: '280-120',
     disableSwipe: true,
   },
+  /* eslint-disable max-len */
+  // TODO
+  // gaa: This needs to be false because jsdom is providing a mocked DOM and react-lazyload
+  // manipulates the DOM so it crashes when attempting to create placeholders
+  // I'll need to look into this further later to see if we can fix this.
+  lazyLoad: false,
+  /* eslint-enable max-len */
 }
 
 describe('ListingCarousel', () => {
--- a/packages/react-ui-ag/src/Listings/MobileMapListing.js
+++ b/packages/react-ui-ag/src/Listings/MobileMapListing.js
@@ -15,6 +15,13 @@ const buttonPropTypes = PropTypes.shape({
   valueLocation: PropTypes.string,
 })
 
+const REACT_LAZYLOAD_PROP_DEFAULTS = {
+  offset: [250, -100],
+  resize: true,
+  width: 280,
+  height: 120,
+}
+
 @themed(/^MobileMapListing/)
 export default class MobileMapListing extends Component {
   static propTypes = {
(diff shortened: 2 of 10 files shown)


nray621:
LGTM 👍

packages/react-ui-ag/src/Listings/MobileMapListing.js
```diff
@@ -15,7 +15,14 @@ const buttonPropTypes = PropTypes.shape({
...
 
-@themed(/^MobileMapListing/)
+const REACT_LAZYLOAD_PROP_DEFAULTS = {
+  offset: [250, -100],
+  resize: true,
+  width: 280,
+  height: 120,
+}
+
+@themed(/^MobileMapListing/, { pure: true })
```
bethstevic: This isn't a pure component so you should remove `{ pure: true }`

bethstevic:
Other than that one thing looks good 👍

bethstevic:
Sorry lol, didn't mean to close your PR

packages/react-ui-ag/src/Carousels/__tests__/ListingCarousel-test.js
```diff
@@ -176,6 +176,13 @@ const listingProps = {
     dimensions: '280-120',
     disableSwipe: true,
   },
+  /* eslint-disable max-len */
+  // TODO
+  // gaa: This needs to be false because jsdom is providing a mocked DOM and react-lazyload
+  // manipulates the DOM so it crashes when attempting to create placeholders
+  // I'll need to look into this further later to see if we can fix this.
+  lazyLoad: false,
```
tadjohnston: This is fine. We've had to do this on all places we've used this lib for the same reason 😄

packages/react-ui-ag/src/Listings/MobileMapListing.js
grenti: Done. #mergingconflictsfrommasterbranchishard :(

packages/react-ui-ag/src/Carousels/__tests__/ListingCarousel-test.js
grenti: Ok I'll remove the comment then 👍

packages/react-ui-ag/src/Listings/MobileMapListing.js
```diff
@@ -142,6 +154,30 @@ export default class MobileMapListing extends Component {
     )
   }
 
+  renderPhotoCarousel() {
+    const { theme, lazyLoad, photos } = this.props
+    const photoProps = {
+      showNav: true,
+      ...photos,
```
tadjohnston: Did you mean to spread these here?

grenti: Good catch - I only meant to spread it below.

packages/react-ui-ag/src/Listings/MobileMapListing.js
```diff
@@ -142,6 +154,30 @@ export default class MobileMapListing extends Component {
...
+    }
+
+    return (
+      <div className={theme.MobileMapListing_Top}>
+        <ListingComponents.Photos
+          {...photoProps}
+          className={theme.MobileMapListing_Photos}
+        />
+      </div>
+    )
```
tadjohnston: I think a cleaner approach would be something like this IMO.  Thoughts?
```
const { theme, photos } = this.props
let lazyLoad = this.props.lazyLoad

if (lazyLoad && typeof lazyLoad === 'boolean') {
  lazyLoad = REACT_LAZYLOAD_PROP_DEFAULTS
}

return (
  <div className={theme.MobileMapListing_Top}>
    <ListingComponents.Photos
      showNav
      photos={photos}
      lazyLoad={lazyLoad}
      className={theme.MobileMapListing_Photos}
    />
  </div>
)

```

Also, is there a way to remove the outer (theme.MobileMapListing_Top) div?  Maybe pass the className directly to `ListingComponents.Photos`?

packages/react-ui-ag/src/Carousels/__tests__/ListingCarousel-test.js
tadjohnston: I thought we made the carousel `lazyLoad` prop false by default though... weird

packages/react-ui-ag/src/Listings/MobileMapListing.js
grenti: @tadjohnston so apparently I should have keep the spread - it was just recently changed in the ListingComponents Photo component (and getting spread there as well in the destructuring assignment).

tadjohnston: I'm confused. If it's spread on the `ListingComponents.Photo`, why would you spread it here too?

packages/react-ui-ag/src/Listings/MobileMapListing.js
grenti: Much cleaner. Thanks I'll update this. However, I think we need to refactor ListingComponents or actually how we distribute props in ListingCarousel. It's kinda confusing what's getting destructured because of duplicate prop names etc.

packages/react-ui-ag/src/Listings/MobileMapListing.js
grenti: So I think we need to refactor how we are using and spreading props from ListingCarousel down to the ListingComponents. There are duplicate prop names and levels of sharing that make it confusing to know exactly what's getting passed around. It'll be easier if I show you.

PR #200, opened 2018-03-01
fix(Desktop Hybrid Map): support selected class to ListItem, List, Menu, and DropdownMenu

mneumark:

[Story](https://rentpath.leankit.com/card/628904109)
* add `selected` bool to `ListItem` that triggers CSS class
* add `selectedIndex` prop to `List`
* add `selectedIndex` prop to `Menu`
* pass `selectedIndex` prop from `DropdownMenu` to `Menu`
* added support for `ListItem-selected` to GridViewHeader CSS

### DIFF
--- a/packages/react-ui-core/src/List/List.js
+++ b/packages/react-ui-core/src/List/List.js
@@ -21,6 +21,7 @@ export default class List extends PureComponent {
     items: PropTypes.array,
     orientation: PropTypes.string,
     highlightIndex: PropTypes.number,
+    selectedIndex: PropTypes.number,
   }
 
   static defaultProps = {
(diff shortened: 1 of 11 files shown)


nray621:
LGTM 👍

packages/react-ui-core/src/List/List.js
```diff
@@ -78,6 +80,7 @@ export default class List extends PureComponent {
           <Item
             {...itemProps}
             highlight={highlightIndex === i}
+            selected={selectedIndex === i}
```
tadjohnston: What's the difference between these 2?

tadjohnston:
I'm pretty sure the reason for the `highlight` prop was to be vague so you _could_ use it for a selected index.

PR #205, opened 2018-03-03
feat(Hybrid Map): fixed LazyLoading forceCheck and Listing performance

grenti:

[Card](https://rentpath.leankit.com/card/624494614)
- [x] Fixed react-lazyload forcecheck issue
during onSlide
- [x] Added feature to MobileMapListing to only render PhotoCarousel when active
else display background image photo
- [x] Updated MobileMapListing.css for Photo to display
background Photo with appropriate height

### DIFF
--- a/packages/react-ui-ag/src/Listings/MobileMapListing.js
+++ b/packages/react-ui-ag/src/Listings/MobileMapListing.js
@@ -50,9 +50,18 @@ export default class MobileMapListing extends Component {
     photos: {},
   }
 
+  constructor(props) {
+    super(props)
+    this.loadedCarousel = false
+  }
+
   shouldComponentUpdate(nextProps) {
     return this.props.isActive !== nextProps.isActive ||
       this.props.listing.id !== nextProps.listing.id
+  }
+
+  get listingPhotos() {
+    return get(this.props.listing, 'photos', [])
   }
 
   @autobind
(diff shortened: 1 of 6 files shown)


packages/react-ui-ag/src/Listings/MobileMapListing.js
```diff
@@ -99,6 +108,11 @@ export default class MobileMapListing extends Component {
     }
   }
 
+  @autobind
+  handlePhotoCarouselSlide(index) { // eslint-disable-line no-unused-vars
```
nray621: I'm guessing there is a reason, but I'm confused why you're giving this function a parameter at all

nray621:
Other than the one comment above, LGTM 👍

grenti:
@nray621 - A grandchild component `Carousel`, is using the handler and it's signature expects an `index` param. Even though we aren't using it here, I want to make it clear that an index is passed into the callback so that future users don't need to dig three levels deep into `<ListComponents.Photos />`, `<PhotoCarousel /> and finally `<Carousel />` to find out what the actual signature of the prop func call should be. 

This is similar to Express.js middleware callback of `(res, req, next) => { }`. Even if you don't use `req` or `res`, you always pass those params since the caller (Express) expects them. Just trying to make clear intent by using some sort of convention since there's a long chain. 

So... what say you mate? 
[] Yay!
[] Nay!!

nray621:
Ahh gotcha. Yeah I figured it was something along those lines. I'm fine either way

bethstevic:
Looks good 👍

PR #207, opened 2018-03-08
Lazyloading single Photo in MobileMapListing

grenti:

[Card](https://rentpath.leankit.com/card/634304886)
* [x] Lazyload Photo on MobileMapListing when listing is inactive and carousel not loaded
* [x] Forcecheck lazyload library when isActive prop changes

### DIFF
--- a/packages/react-ui-ag/src/Listings/MobileMapListing.js
+++ b/packages/react-ui-ag/src/Listings/MobileMapListing.js
@@ -4,6 +4,7 @@ import themed from 'react-themed'
 import classnames from 'classnames'
 import autobind from 'autobind-decorator'
 import get from 'lodash/get'
+import LazyLoad, { forceCheck } from 'react-lazyload'
 import { Button, ToggleButton, ListingComponents, ListingCell } from '@rentpath/react-ui-core'
 import { Banner } from '../Banners'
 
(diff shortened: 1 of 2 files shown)


packages/react-ui-ag/src/Listings/MobileMapListing.js
```diff
@@ -55,6 +56,12 @@ export default class MobileMapListing extends Component {
     this.loadedCarousel = false
   }
 
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.isActive !== this.props.isActive) {
```
tadjohnston: So this checks only 2 things to forceCheck:
1. active -> inactive
2. inactive -> active

Does that really need to happen? Are you sure this is really doing anything?

grenti: When I check when it moves in either state it makes sure that the Photo is loaded when the Carousel slides the next three listings. If I don't do a forceCheck here, the lazyloaded placeholder always stays in place regardless of the onSlide call in the parent ListingCarousel's onSlide event.

grenti:
@tadjohnston so I refactored the logic here - so only if `nextProps.isActive` changes and it is true then `forceCheck`.

PR #209, opened 2018-03-14
feat(AutoSuggestField Menu): Fixes debounce issue and menu

yasabere:

[Card](https://rentpath.leankit.com/card/634890555)
- Added ability to include non-clickable options in suggestions
- Added tests
- Options are disabled by adding 'disabled:true' attribute to option object

### DIFF
--- a/packages/react-ui-core/src/Menu/Menu.js
+++ b/packages/react-ui-core/src/Menu/Menu.js
@@ -114,10 +114,10 @@ export default class Menu extends PureComponent {
   handleSelection() {
     const { onItemSelect } = this.props
 
-    const value = this.highlightedOption
+    const option = this.highlightedOption
 
-    if (onItemSelect && value && !value.disabled) {
-      onItemSelect(value, this.state.highlightIndex)
+    if (onItemSelect && option && !option.disabled) {
+      onItemSelect(option, this.state.highlightIndex)
     }
   }
 
--- a/packages/react-ui-core/src/Menu/Menu.js
+++ b/packages/react-ui-core/src/Menu/Menu.js
@@ -47,14 +47,15 @@ export default class Menu extends PureComponent {
   constructor(props) {
     super(props)
     this.state = {
-      highlightIndex: props.highlightIndex || 0,
-      indexedOptionIndex: props.highlightIndex || 0,
+      highlightIndex: -1,
+      indexedOptionIndex: -1,
       indexedOptions: this.generateIndexedOptions(),
     }
   }
 
   componentDidMount() {
     document.addEventListener('keydown', this.onKeyDown)
+    this.highlightIndexedOption(0)
   }
 
   componentWillReceiveProps(nextProps) {
@@ -65,7 +66,7 @@ export default class Menu extends PureComponent {
     if (this.props.options !== nextProps.options) {
       this.setState({
         indexedOptions: this.generateIndexedOptions(),
-      })
+      }, this.highlightIndexedOption(0))
     }
   }
 
--- a/packages/react-ui-core/src/Menu/Menu.js
+++ b/packages/react-ui-core/src/Menu/Menu.js
@@ -46,7 +46,6 @@ export default class Menu extends PureComponent {
 
   constructor(props) {
     super(props)
-
     this.state = {
       highlightIndex: props.highlightIndex || 0,
       indexedOptionIndex: props.highlightIndex || 0,
@@ -107,7 +106,7 @@ export default class Menu extends PureComponent {
     )
   }
 
-  get option() {
+  get highlightedOption() {
     return this.options[this.state.highlightIndex || 0]
   }
 
@@ -115,7 +114,7 @@ export default class Menu extends PureComponent {
   handleSelection() {
     const { onItemSelect } = this.props
 
-    const value = this.option
+    const value = this.highlightedOption
 
     if (onItemSelect && value && !value.disabled) {
       onItemSelect(value, this.state.highlightIndex)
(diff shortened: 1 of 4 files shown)


packages/react-ui-core/src/Menu/Menu.js
```diff
@@ -93,28 +95,26 @@ export default class Menu extends PureComponent {
...
-
   get options() {
     return this.props.options || []
   }
 
   get items() {
-    return this.options.map(option => (
-      typeof option === 'object' ? option.label : option
-    ))
+    return this.options.map(option => ((typeof option === 'object' && option.label !== undefined) ? option.label : option))
```
mneumark: I think this would be more readable if the ternary were on a separate line like it was before.

packages/react-ui-core/src/Menu/Menu.js
```diff
@@ -46,6 +47,7 @@ export default class Menu extends PureComponent {
   constructor(props) {
     super(props)
 
+    this.state = {}
```
bethstevic: Why are you setting state as an empty object here and assigning it a value right below?

packages/react-ui-core/src/Menu/Menu.js
```diff
@@ -93,28 +95,26 @@ export default class Menu extends PureComponent {
...
   }
 
   get items() {
-    return this.options.map(option => (
-      typeof option === 'object' ? option.label : option
-    ))
+    return this.options.map(option => ((typeof option === 'object' && option.label !== undefined) ? option.label : option))
+  }
+
+  option() {
```
mneumark: What do you think about renaming this to `highlightedOption` to make it more clear? Also I see no reason this shouldn't be a getter.

packages/react-ui-core/src/Menu/Menu.js
```diff
@@ -93,28 +95,26 @@ export default class Menu extends PureComponent {
...
+    return this.options[this.state.highlightIndex || 0]
   }
 
   @autobind
   handleSelection() {
-    const { onItemSelect, options } = this.props
+    const { onItemSelect } = this.props
 
-    const option = options[this.state.highlightIndex]
+    const value = this.option()
```
mneumark: Why did you rename this to value when you went out of your way to rename the function away from value to make it more clear?

packages/react-ui-core/src/Menu/__tests__/Menu-test.js
```diff
@@ -35,6 +35,8 @@ const objectOptionsWithLabel = [
   },
   {
     label: 'Baz label',
+    value: 'Baz Value',
+    disabled: true,
```
mneumark: Do we need some tests for validating what happens when you disable something?

packages/react-ui-core/src/Menu/Menu.js
```diff
@@ -93,28 +95,26 @@ export default class Menu extends PureComponent {
...
     return this.props.options || []
   }
 
   get items() {
-    return this.options.map(option => (
-      typeof option === 'object' ? option.label : option
-    ))
+    return this.options.map(option => ((typeof option === 'object' && option.label !== undefined) ? option.label : option))
+  }
+
```
bethstevic: Can we break this into multiple lines? Like this:
```
return this.options.map(option => 
       ((typeof option === 'object' && option.label !== undefined) ? 
             option.label :
             option
        )
     )
```

packages/react-ui-core/src/Menu/__tests__/Menu-test.js
tadjohnston: Yes!

packages/react-ui-core/src/Menu/Menu.js
tadjohnston: Some of this was me having these renamed. I don't think it was ever finished...

packages/react-ui-core/src/Menu/__tests__/Menu-test.js
yasabere: There is a test for that already but the description might not be clear enough

packages/react-ui-core/src/Menu/Menu.js
yasabere: Fixed

packages/react-ui-core/src/Menu/Menu.js
yasabere: Fixed

packages/react-ui-core/src/Menu/__tests__/Menu-test.js
yasabere: Fixed

mneumark: Okay, to be clear, he wrote tests in the PR from 2 days ago but had issues with that resolution and in this PR he is fixing stuff but the tests from before remain.

packages/react-ui-core/src/Menu/Menu.js
nray621: I'm still confused why this is `value` rather than `option`. If it is named `value` then we end up with an object named `value` that has a `value` attribute

packages/react-ui-core/src/Menu/Menu.js
```diff
@@ -93,28 +93,31 @@ export default class Menu extends PureComponent {
...
+    return this.options[this.state.highlightIndex || 0]
   }
 
   @autobind
   handleSelection() {
-    const { onItemSelect, options } = this.props
+    const { onItemSelect } = this.props
 
-    const option = options[this.state.highlightIndex]
+    const value = this.highlightedOption
```
nray621: I'm still confused why this is `value` rather than `option`. If it is named `value` then we end up with an object named `value` that has a `value` attribute

nray621:
LGTM so far 👍

nray621:
👍

PR #211, opened 2018-03-16
634980824 Prevent modal from closing after dragging mouse outside

barelyfitz:
LeanKit card (rent-js): [Prevent modal closing after dragging mouse](https://rentpath.leankit.com/card/634980824)

In some cases, especially when the modal contains a draggable control such as a price slider, the user might click inside the modal, hold the button to drag, then release the button outside of the modal. Currently this gets registered as a click on the overlay, and the modal closes, which is not what the user intended.

I examined some third-party modal react components and copied a technique for verifying the click on the overlay actually started and ended on the overlay.

Changes:
* Adds event handlers to the Overlay component to prevent triggering the onClick handler in cases where the overlay wasn't intentionally clicked.
* Modify unit tests that broke because of these changes
* Add new unit tests to simulate new use cases

How to test:
* View a modal.
* Move mouse cursor outside the modal, click the mouse button.
  - Verify that the close modal action occurs.
* Move mouse cursor into the modal, press and hold the mouse button, move the mouse cursor outside the modal, release the mouse button.
  - Verify the close modal action does not occur.
* Move mouse cursor outside the modal, press and hold the mouse button, move the mouse cursor inside the modal, release the mouse button.
  - Verify the close modal action does not occur.
* Move mouse cursor inside the modal, click the mouse button.
  - Verify the close modal action does not occur.

### DIFF
--- a/packages/react-ui-core/src/Modal/Overlay.js
+++ b/packages/react-ui-core/src/Modal/Overlay.js
@@ -10,6 +10,8 @@ import themed from 'react-themed'
 export default class Overlay extends PureComponent {
   static propTypes = {
     className: PropTypes.string,
+    onMouseDown: PropTypes.func,
+    onMouseUp: PropTypes.func,
     onClick: PropTypes.func,
     theme: PropTypes.object,
     children: PropTypes.node,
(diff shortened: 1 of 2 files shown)


packages/react-ui-core/src/Modal/Overlay.js
```diff
@@ -23,9 +23,29 @@ export default class Overlay extends PureComponent {
...
+  handleMouseDown(e) {
+    // Verify that the click started on the overlay
+    this.clickedOutside = (e.target === this.overlay)
+  }
+
+  handleMouseUp(e) {
+    // Verify that the click ended on the overlay
+    if (e.target !== this.overlay) {
+      this.clickedOutside = false
+    }
```
lfender6445: could we add callbacks? (`if (this.props.handleMouseUp) this.props.handleMouseUp()`)

lfender6445:
👍

packages/react-ui-core/src/Modal/Overlay.js
barelyfitz: @lfender6445 Added `onMouseDown` and `onMouseUp` handlers to pass through these events just in case user wants to use them.

PR #212, opened 2018-03-19
feat(AutoSuggestField): adds onVisibilityChange prop to AutoSuggestFi…

yasabere:
…eld so the state can be passed

[Card](https://rentpath.leankit.com/card/637409410)
-Create function in autosuggestfield to return
visible state


bethstevic:
Looks good 👍

nray621:
LGTM 👍

PR #213, opened 2018-03-20
fix(AutoSuggestField): autoSuggestField changed to pass 'value' prop …

yasabere:
…to the state when it is updated

[Card](https://rentpath.leankit.com/card/638551669)
- AutoSuggestField changed to pass 'value' prop to the state when it is updated
- Added a componentWillRecieveProps to update 'value' in state when the prop changes
- Modifed dropdown to call onvisibilitychange on handleDocumentClick

Fixes Bugs
- react-ui-core autosuggestfield: Value prop is not passed to state on update
- react-ui-core dropdown: onvisibilitychange is not called on handleDocumentClick

### DIFF
--- a/packages/react-ui-core/src/AutoSuggestField/AutoSuggestField.js
+++ b/packages/react-ui-core/src/AutoSuggestField/AutoSuggestField.js
@@ -81,6 +81,14 @@ export default class AutoSuggestField extends Component {
     document.addEventListener('keydown', this.onKeyDown)
   }
 
+  componentWillReceiveProps(nextProps) {
+    if (!(this.props.value === nextProps.value)) {
+      this.setState({
+        value: nextProps.value,
+      })
+    }
+  }
+
   componentWillUnmount() {
     document.removeEventListener('keydown', this.onKeyDown)
   }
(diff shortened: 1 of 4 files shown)


packages/react-ui-core/src/AutoSuggestField/AutoSuggestField.js
```diff
@@ -81,6 +81,14 @@ export default class AutoSuggestField extends Component {
     document.addEventListener('keydown', this.onKeyDown)
   }
 
+  componentWillReceiveProps(nextProps) {
+    if (!(this.props.value === nextProps.value)) {
```
grenti: I would write a test to check the state once the props are changed.

packages/react-ui-core/src/Dropdown/Dropdown.js
```diff
@@ -70,6 +70,7 @@ export default class Dropdown extends Component {
   handleDocumentClick(event) {
     if (this.state.visible && !this.dropdown.contains(event.target)) {
       this.setState({ visible: false })
+      this.props.onVisibilityChange(false)
```
grenti: I would update the test for `handleDocumentClick` too to check that the `onVisibilityChange` mocked func prop gets called too

packages/react-ui-core/src/AutoSuggestField/AutoSuggestField.js
yasabere: Added

packages/react-ui-core/src/Dropdown/Dropdown.js
yasabere: Added

nray621:
👍

PR #214, opened 2018-03-20
fix(Menu): pass properties (other than label) to Menu ListItems

mpelzsherman:

[Card](https://rentpath.leankit.com/card/591034222)

The Menu component wasn't passing additional props to its ListItems.
This prevented us from adding tagging attributes, etc.

### DIFF
--- a/packages/react-ui-core/src/List/List.js
+++ b/packages/react-ui-core/src/List/List.js
@@ -18,7 +18,13 @@ export default class List extends PureComponent {
       PropTypes.func,
       PropTypes.object,
     ]),
-    items: PropTypes.array,
+    items: PropTypes.arrayOf(
+      PropTypes.oneOfType([
+        PropTypes.node,
+        PropTypes.func,
+        PropTypes.shape({
+          label: PropTypes.node.isRequired,
+        })])),
     orientation: PropTypes.string,
     highlightIndex: PropTypes.number,
     selectedIndex: PropTypes.number,
@@ -49,8 +55,18 @@ export default class List extends PureComponent {
     return `${this.id}-${index}`
   }
 
-  renderListItem(listItem) {
+  itemComponentAndProps(listItem) {
     return parseArgs(listItem, DefaultListItem)
+  }
+
+  renderItem(item) {
+    if (typeof item === 'object') {
+      const { label, ...props } = item
+      return [label, props]
+    } else if (typeof item === 'function') {
+      return [item()]
+    }
+    return [item]
   }
 
   render() {
@@ -66,7 +82,7 @@ export default class List extends PureComponent {
       ...props
     } = this.props
 
-    const [Item, itemProps] = this.renderListItem(listItem)
+    const [Item, baseProps] = this.itemComponentAndProps(listItem)
 
     return (
       <NodeType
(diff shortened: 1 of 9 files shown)


packages/react-ui-core/src/List/List.js
```diff
@@ -76,19 +90,23 @@ export default class List extends PureComponent {
...
-        ))}
+        {items.map((item, i) => {
+          const [children, itemProps] = this.renderItem(item)
+          return (
+            <Item
+              {...itemMetaProps}
+              data-tid={`list-item-${i}`}
+              {...props}
+              {...itemProps}
+              highlight={highlightIndex === i}
```
bethstevic: I don't remember well what itemProps and props contain, but it was set up before so that props would override item props if there were duplicates and this sets it up the opposite way, was that on purpose?

yasabere:
This change my break my pr https://github.com/rentpath/ag.js/pull/3932

bethstevic:
Besides the one comment (if it was intentional) looks good to me :+1:

packages/react-ui-core/src/List/List.js
```diff
@@ -49,10 +55,18 @@ export default class List extends PureComponent {
     return `${this.id}-${index}`
   }
 
-  renderListItem(listItem) {
+  itemMetaProps(listItem) {
     return parseArgs(listItem, DefaultListItem)
```
lfender6445: what are itemMetaProps?

packages/react-ui-core/src/List/__tests__/List-test.js
```diff
@@ -27,6 +50,24 @@ describe('List', () => {
...
+    it('passes through object props', () => {
+      const snap = renderer
+        .create(<List items={menuItems} />)
+        .toJSON()
+      expect(snap).toMatchSnapshot()
+    })
+
+    it('handles labelless items', () => {
+      const errorFunc = global.console.error
+      global.console = { error: jest.fn() }
```
lfender6445: will this wipe out console for all other tests?

packages/react-ui-core/src/List/List.js
```diff
@@ -18,7 +18,13 @@ export default class List extends PureComponent {
       PropTypes.func,
       PropTypes.object,
     ]),
-    items: PropTypes.array,
+    items: PropTypes.arrayOf(
+      PropTypes.oneOfType([
+        PropTypes.node,
+        PropTypes.func,
```
lfender6445: is there existing test case for when items are functions?

packages/react-ui-core/src/Menu/Menu.js
```diff
@@ -173,7 +164,7 @@ export default class Menu extends PureComponent {
           className,
           theme.Menu
         )}
-        items={this.items}
+        items={options}
```
lfender6445: are you sure this shouldn't be `this.options` ?

packages/react-ui-core/src/List/List.js
tadjohnston: Yeah, this was on purpose. I wasn't seeing a reason why we were able to override these.

packages/react-ui-core/src/List/__tests__/List-test.js
tadjohnston: No, because it gets reset after [here](https://github.com/jest-community/eslint-plugin-jest#usage).  We couldn't find an easier way to do this.  Open to suggestions.

packages/react-ui-core/src/List/List.js
tadjohnston: Yes, that could be `() => (<div>foo</foo>)`.

packages/react-ui-core/src/Menu/Menu.js
tadjohnston: Good catch! Yes it should be 😄

packages/react-ui-core/src/List/__tests__/List-test.js
lfender6445: yeah but you're taking out console.log and otheres - what about 

```
  let original

  beforeEach(() => {
    original = console.error
    console.error = jest.fn()
    console.error('you cant see me')
  })

  afterEach(() => {
    console.log('log still works')
    console.error('you cant see me')
    console.error.mockClear()
    console.error = original
    console.error('now you can')
  })

```

tadjohnston: We don't need to before each them... I see the issues though, it should probably be:
```
global.console.error = jest.fn()
```

packages/react-ui-core/src/List/List.js
tadjohnston: This should probably be renamed since it's returning both a component and props.

packages/react-ui-core/src/List/List.js
mpelzsherman: Good catch @lfender6445 ! Test added - and bug fixed!

PR #217, opened 2018-03-22
feat(Property Link Email): create EmailModal component

mneumark:

[Story](https://rentpath.leankit.com/card/636455802)

* use state for `isOpen` instead of the props. this is so it can be managed by `onClose` better.
* create EmailModal
* fix bug causing listing.isFavorited knob to show up everywhere
* group core css files in AG base
* switch Overlay to autobind decorators
* allow props through to the close button in the Modal Body


packages/react-ui-ag/src/Modals/EmailModal.js
```diff
@@ -0,0 +1,77 @@
...
+  static defaultProps = {
+    theme: {},
+  }
+
+  @autobind
+  onSubmit(event, fields) {
+    const { onSubmit } = this.props
+
+    if (onSubmit) onSubmit(event, fields)
+  }
```
barelyfitz: Is there any need to have `this.onSubmit` if all it does is turnaround and call the `this.props.onSubmit` with the same arguments it receives? Why not just use `onSubmit={this.props.onSubmit}` down in the render function?

mneumark: Good point, I expected I'd need to add something and never did.

mneumark: Fixed.

packages/react-ui-ag/src/Modals/__tests__/EmailModal-test.js
```diff
@@ -0,0 +1,60 @@
...
+    theme,
+  }
+  it('does not render the error message when no error is passed', () => {
+    const wrapper = shallow(<EmailModal />)
+    expect(wrapper.find('[data-tid="error"]').exists()).toBeFalsy()
+  })
+
+  it('renders the error message when there is an error passed', () => {
+    const wrapper = shallow(<EmailModal error="foo bar baz" />)
+    expect(wrapper.find('[data-tid="error"]').exists()).toBeTruthy()
```
barelyfitz: expect text to be "foo bar baz" ?

packages/react-ui-ag/src/Modals/__tests__/EmailModal-test.js
```diff
@@ -0,0 +1,60 @@
...
+  })
+
+  it('does not render the footer when no footer is passed', () => {
+    const wrapper = shallow(<EmailModal />)
+    expect(wrapper.find('[data-tid="footer"]').exists()).toBeFalsy()
+  })
+
+  it('renders the footer when a footer is passed', () => {
+    const wrapper = shallow(<EmailModal footer="hello there" />)
+    expect(wrapper.find('[data-tid="footer"]').exists()).toBeTruthy()
```
barelyfitz: Expect text to be "hello there" ?

packages/react-ui-ag/src/Modals/__tests__/EmailModal-test.js
```diff
@@ -0,0 +1,60 @@
...
+  })
+})
```
barelyfitz: Add same tests for `header` as for `footer`?

packages/react-ui-ag/src/Modals/EmailModal.js
barelyfitz: Also my preference is to name all handlers starting with "handle", so `handleSubmit` in this case unless you decide to remove this.

packages/react-ui-core/src/Modal/ClosableModal.js
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  shouldComponentUpdate(nextProps, nextState) {
+    return nextProps.isOpen !== this.props.isOpen ||
+      nextProps.children !== this.props.children ||
+      nextState.isOpen !== this.state.isOpen
+  }
+
+  @autobind
+  closeModal() {
```
barelyfitz: My preference is to name all handlers starting with "handle", so `handleClose` or `handleCloseModal` in this case.

packages/react-ui-ag/src/Modals/__tests__/EmailModal-test.js
barelyfitz: Add test `it('allows other props through ...'` ?

packages/react-ui-core/src/Modal/ClosableModal.js
```diff
@@ -0,0 +1,72 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+import themed from 'react-themed'
+import classnames from 'classnames'
+import autobind from 'autobind-decorator'
+
+import { Modal } from '../Modal'
+
+@themed(/^ClosableModal/)
+export default class ClosableModal extends Component {
```
tadjohnston: I'm not getting the purpose of this component. Why can't this already be done in the existing Modal?

packages/react-ui-core/src/Modal/ClosableModal.js
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.isOpen !== this.props.isOpen) {
+      this.setState({ isOpen: nextProps.isOpen })
+    }
+  }
+
+  shouldComponentUpdate(nextProps, nextState) {
+    return nextProps.isOpen !== this.props.isOpen ||
```
tadjohnston: So basically anything here will return truthy?

packages/react-ui-core/src/Modal/__tests__/Modal-test.js
```diff
@@ -20,25 +20,25 @@ describe('Modal/Modal', () => {
       wrapper = shallow(<Modal theme={theme} onClose={onClose} />)
     })
 
-    test('adds `className` prop to the root node', () => {
```
tadjohnston: Not really needed to change this here as they're both synonymous.  Either way it's fine for me though.

packages/react-ui-ag/src/Modals/__tests__/EmailModal-test.js
mneumark: I suppose I could test that.

packages/react-ui-ag/src/Modals/__tests__/EmailModal-test.js
mneumark: I suppose I could test that.

stories/theme/ag/large/Modals/EmailModal.css
```diff
@@ -0,0 +1,110 @@
...
+}
+
+.EmailModal_Header {
+  width: 240px;
+  padding-bottom: 15px;
+}
+
+.EmailModal h3 {
+    margin: 0;
+    padding: 30px 0 15px;
```
barelyfitz: too much indent

packages/react-ui-ag/src/Modals/__tests__/EmailModal-test.js
mneumark: Good catch.

packages/react-ui-core/src/Modal/ClosableModal.js
mneumark: Okay.

packages/react-ui-core/src/Modal/ClosableModal.js
mneumark: No, if it was set to true and it state true, it would return false...

packages/react-ui-core/src/Modal/ClosableModal.js
mneumark: Modal does not have state for `isOpen` and we are setting ourselves up to implement state of isOpen in many of our uses of `Modal`.

packages/react-ui-core/src/Modal/__tests__/Modal-test.js
mneumark: More a matter of standardizing as everywhere else uses `it`.

packages/react-ui-core/src/Modal/ClosableModal.js
tadjohnston: So you're updating when it's either open or closed? I guess I'm not seeing why you just wouldn't make this a PureComponent instead.  You're pretty much doing the same thing.

mneumark: Fair enough. I think making this a pure component could be better.

packages/react-ui-core/src/Modal/ClosableModal.js
tadjohnston: I'd rather have this functionality in `Modal` TBH.  This extra component is just going to confuse people.

tadjohnston: The only check you'd have to add would be from [this line](https://github.com/rentpath/react-ui/pull/217/files#diff-53fc33bc9837831b8b065721991107dbR42): 

```
if (this.props.onClose) {
  this.setState({  isOpen: false })
  this.props.onClose()
}
```

packages/react-ui-core/src/Modal/Modal.js
```diff
@@ -26,9 +27,26 @@ export default class Modal extends PureComponent {
...
   get overlayClose() {
-    const { onClose, closeOnOverlayClick } = this.props
-    return closeOnOverlayClick ? onClose : undefined
+    const { closeOnOverlayClick } = this.props
+    return closeOnOverlayClick ? this.handleClose : undefined
+  }
+
+  @autobind
+  handleClose() {
+    this.setState({ isOpen: false })
```
tadjohnston: This needs to be inside of the `if (this.props.onClose` check

tadjohnston: Hmm.. I guess it really doesn't matter.  I

packages/react-ui-ag/src/Modals/EmailModal.js
```diff
@@ -0,0 +1,69 @@
...
+          <Field
+            name="email"
+            label={label}
+            defaultValue={email}
+            key={email}
+            {...field}
+          />
+          {error && <span className={theme.EmailModal_ErrorMessage} data-tid="error">{error}</span>}
+          <div className={theme.EmailModal_Submit}>
+            <Button type="submit" name="submit" {...submitButton}>Submit</Button>
```
barelyfitz: If `submitButton` is passing in Button props, perhaps rename `submitButtonProps` for more clarity.

Consider adding a prop 'submitButtonText' and default to string 'Submit'.

tadjohnston:
Agree with @barelyfitz, nice work!

tadjohnston:
@barelyfitz I'm suggesting this to @mneumark:
```

tadjohnston:
@barelyfitz is cool with the approach @mneumark. Merge away!

PR #218, opened 2018-03-22
feat(googlemap): addition of google map

tadjohnston:

[Card](https://rentpath.leankit.com/card/636402673)
Adding google map components with basic
functionality

### DIFF
--- a/packages/react-ui-core/src/Gmap/Gmap.js
+++ b/packages/react-ui-core/src/Gmap/Gmap.js
@@ -1,0 +1,67 @@
+import React, { PureComponent } from 'react'
+import PropTypes from 'prop-types'
+import themed from 'react-themed'
+import { parseArgs } from '@rentpath/react-ui-utils'
+import GmapBase from './GmapBase'
+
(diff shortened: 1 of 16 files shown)


packages/react-ui-core/src/Gmap/Gmap.js
```diff
@@ -0,0 +1,63 @@
...
+    return createComponent(loadingElement, theme.Gmap_Spinner)
+  }
+
+  render() {
+    return (
+      <GmapBase
+        {...this.props}
+        containerElement={this.containerElement}
+        mapElement={this.mapElement}
+        loadingElement={this.loadingElement}
```
ninomichaels: might be useful to comment that these are props for `withScriptjs` and `withGoogleMap`, looks weird that you don't do anything with them in `GmapBase`

tadjohnston: Yeah, I really dislike the HOC this library uses for these.  There's no way to get rid of them.

PR #220, opened 2018-03-26
Added LinkList component

paudyalhari:


packages/react-ui-core/package.json
```diff
@@ -26,6 +26,7 @@
     ]
   },
   "devDependencies": {
+    "@rentpath/react-redux-router": "^3.0.2",
```
tadjohnston: This shouldn't be here in react-ui-core at all.

packages/react-ui-core/src/LinkList/LinkItem.js
```diff
@@ -0,0 +1,48 @@
...
+        {text}
+      </Link>
+    )
+  }
+
+  render() {
+    const { className, tagItem, dataTestId } = this.props
+
+    return (
+      <li className={className} data-test-id={dataTestId} data-tag_item={tagItem}>
```
tadjohnston: We use `data-tid` and not `data-test-id`.

packages/react-ui-core/src/LinkList/LinkItem.js
```diff
@@ -0,0 +1,48 @@
...
+    text: PropTypes.string.isRequired,
+    className: PropTypes.string,
+    tagItem: PropTypes.string,
+    useAnchorTag: PropTypes.bool,
+    title: PropTypes.string,
+    dataTestId: PropTypes.string,
+  }
+
+  static defaultProps = {
+    tagItem: 'link',
```
tadjohnston: Please do not pass data tags like as props like this.

packages/react-ui-core/src/LinkList/LinkList.js
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  render() {
+    const { links, theme, className, useAnchorTag } = this.props
+
+    return (
+      <ul className={classnames(className, theme.list)}>
+        { links.map(item => (
+          <LinkItem
+            dataTestId={this.createDataTestId(item)}
```
tadjohnston: Let's not pass this in directly like this. These should be static ids anyway.

packages/react-ui-core/src/LinkList/LinkList.js
```diff
@@ -0,0 +1,47 @@
...
+            key={item.url}
+            url={item.url}
+            text={item.text}
+            className={classnames(
+              item.className,
+              theme.item,
```
tadjohnston: If you're adding theming, you'll need to wrap this component in `react-themed` `@themed()` and add it the correct way. I can help you if you'd like.

paudyalhari:
Removing this PR since we don't want to have dependency on redux in react-ui.

mpelzsherman:
It is unfortunate that we can't use a `Link` component in `react-core`.

tadjohnston:
@mpelzsherman all you have to do is pass in `nodeType: Link` and it will work as it's supposed to 😄

PR #221, opened 2018-03-28
refactor(google map): using more suitable pkg

tadjohnston:

[Card](https://rentpath.leankit.com/card/636402673)
Switching to a library more suited for our
needs. This one has more flexibility for marker options and interactions


packages/react-ui-core/src/Gmap/Gmap.js
```diff
@@ -1,67 +1,49 @@
...
+    })(GmapBase)
   }
 
-  get mapElement() {
-    const { mapElement, theme } = this.props
-    return createComponent(mapElement, theme.Gmap_Element)
-  }
+  render() {
+    const {
+      apiKey,
```
ninomichaels: do these 4 props need to be here?

packages/react-ui-core/src/Gmap/GmapBase.js
```diff
@@ -1,32 +1,104 @@
...
-    )
+    return [
+      this.spinner,
+      <Map
+        className={classnames(
+          className,
+          theme.Gmap,
+        )}
+        key="map"
+        google={this.props.google}
```
ninomichaels: should this be in the destructuring assignment above? in `propTypes`?

packages/react-ui-core/src/Gmap/Gmap.js
tadjohnston: Yeah because I combined all props into one component so I'd rather not pass those on.

packages/react-ui-core/src/Gmap/GmapBase.js
tadjohnston: Yeah, I probably need to include it there. 👍

tadjohnston:
Not gonna work.

PR #227, opened 2018-03-30
refactor(google map): refactoring components

tadjohnston:

The google map didn't play well with things like SSR and such.  Refactoring to use a generic way of
doing the maps.

### DIFF
--- a/packages/react-ui-core/src/Gmap/Gmap.js
+++ b/packages/react-ui-core/src/Gmap/Gmap.js
@@ -1,67 +1,86 @@
 import React, { PureComponent } from 'react'
 import PropTypes from 'prop-types'
-import themed from 'react-themed'
+import autobind from 'autobind-decorator'
 import { parseArgs } from '@rentpath/react-ui-utils'
 import GmapBase from './GmapBase'
 
-const createComponent = (element, className) => {
-  if (React.isValidElement(element)) return element
-
-  const [Container, props] = parseArgs(element, 'div', { className })
-  return <Container {...props} />
-}
-
-@themed(/^Gmap/, { pure: true })
+const API_BASE_URL = 'https://maps.googleapis.com/maps/api/js'
 
 export default class Gmap extends PureComponent {
   static propTypes = {
-    loadingElement: PropTypes.oneOfType([
-      PropTypes.object,
-      PropTypes.element,
+    apiKey: PropTypes.string.isRequired,
+    libraries: PropTypes.arrayOf(
+      PropTypes.string,
+    ),
+    spinner: PropTypes.oneOfType([
+      PropTypes.func,
       PropTypes.node,
-    ]),
-    containerElement: PropTypes.oneOfType([
       PropTypes.object,
-      PropTypes.element,
     ]),
-    mapElement: PropTypes.oneOfType([
-      PropTypes.object,
-      PropTypes.element,
-    ]),
-    theme: PropTypes.object,
   }
 
   static defaultProps = {
-    theme: {},
+    libraries: [],
   }
 
-  get containerElement() {
-    const { containerElement, theme } = this.props
-    return createComponent(containerElement, theme.Gmap_Container)
+  constructor(props) {
+    super(props)
+    this.state = {
+      loaded: false,
+      spinner: props.spinner,
+    }
   }
 
-  get mapElement() {
-    const { mapElement, theme } = this.props
-    return createComponent(mapElement, theme.Gmap_Element)
+  componentDidMount() {
+    if (this.isInitialized) {
+      this.scriptLoaded()
+      return
+    }
+    this.loadScript()
   }
 
-  get loadingElement() {
-    const { loadingElement, theme } = this.props
-    return createComponent(loadingElement, theme.Gmap_Spinner)
+  get isInitialized() {
+    return !!(window.google && window.google.maps)
+  }
+
+  get spinner() {
+    const { spinner, loaded } = this.state
+    const props = { key: 'map-spinner', loading: !loaded }
+
+    if (spinner) {
+      if (React.isValidElement(spinner)) return React.cloneElement(spinner, props)
+      return React.createElement(...parseArgs(spinner, 'div', props))
+    }
+
+    return null
+  }
+
+  @autobind
+  scriptLoaded() {
+    this.setState({ loaded: true })
+  }
+
+  async loadScript() {
+    const { apiKey, libraries } = this.props
+    const url = `${API_BASE_URL}?key=${apiKey}&libraries=${libraries.join()}`
+
+    await import(/* webpackChunkName: "scriptjs" */ 'scriptjs').then(scriptjs => {
+      scriptjs(url, this.scriptLoaded)
+    })
   }
 
   render() {
-    /* NOTE: containerElement, mapElement and loadingElement props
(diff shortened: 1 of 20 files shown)


packages/react-ui-core/src/Gmap/Gmap.js
```diff
@@ -1,67 +1,87 @@
...
-    mapElement: PropTypes.oneOfType([
-      PropTypes.object,
-      PropTypes.element,
-    ]),
-    theme: PropTypes.object,
   }
 
   static defaultProps = {
-    theme: {},
+    libraries: LIBRARIES,
```
bethstevic: Just curious, why didn't you just set the default libraries to `[]`?

tadjohnston: Good question, I think it was because I actually had something in it originally.  I'll switch it.

bethstevic:
Looks good to me 👍

packages/react-ui-core/src/Gmap/Gmap.js
```diff
@@ -1,67 +1,87 @@
...
   }
 
-  get containerElement() {
-    const { containerElement, theme } = this.props
-    return createComponent(containerElement, theme.Gmap_Container)
+  constructor(props) {
+    super(props)
+    this.state = {
+      loaded: false,
+      spinner: props.spinner,
```
ninomichaels: `this.props.spinner`?

packages/react-ui-core/src/Gmap/Marker.js
```diff
@@ -0,0 +1,63 @@
...
+      this.setupEvents()
+    }
+
+    return null
+  }
+
+  setupEvents() {
+    Object.keys(EVENTS).forEach(name => {
+      if (this.props[name]) {
+        this.marker.addListener(EVENTS[name], this.handleEvent(name))
```
ninomichaels: do we need to worry about tearing these down on unload?

packages/react-ui-core/src/Gmap/Marker.js
```diff
@@ -0,0 +1,63 @@
...
+      <MapContext.Consumer>
+        { context => this.setupMarker(context.map)}
+      </MapContext.Consumer>
+    )
+  }
+}
+
+Marker.propTypes = {}
+
+Object.keys(EVENTS).forEach(name => {
```
ninomichaels: is it worthwhile to memoize `Object.keys(EVENTS)` since you are using on line 35 as well?

packages/react-ui-core/src/Gmap/GmapSpinner.js
```diff
@@ -11,18 +11,44 @@ export default class MapSpinner extends PureComponent {
...
   }
 
+  constructor(props) {
+    super(props)
+    this.state = {
+      loading: this.props.loading,
+    }
+  }
+
+  componentDidUpdate(nextProps) {
```
ninomichaels: did you mean to use a difference lifecycle method here, because 

`componentDidUpdate(prevProps, prevState, snapshot)`

tadjohnston: Here, no. @mneumark asked the same.  Reason being is because `componentWIllReceiveProps` will become deprecated and `getDerivedStateFromProps` won't work how we need it to with a `setTmeout`. I'm open to changing this to something else though...

packages/react-ui-core/src/Gmap/Marker.js
tadjohnston: Hmm... I thought about this and was thinking marker was getting removed on unmount, but it's only getting removed from the map on unmount, so good catch.

packages/react-ui-core/src/Gmap/Gmap.js
tadjohnston: `props` is available in the constructor, either one works.

packages/react-ui-core/src/Gmap/GmapBase.js
```diff
@@ -1,32 +1,105 @@
...
+
+  initMap() {
+    const {
+      zoom,
+      mapControls,
+      center,
+      clickableIcons,
+    } = this.props
+
+    const map = new window.google.maps.Map(this.googleMap.current, {
```
mpelzsherman: Any chance of `window.google.maps` being null here?

packages/react-ui-core/src/Gmap/GmapSpinner.js
ninomichaels: I think calling close in `componentDidUpdate` is the right thing to do, just calling the param `nextProps` is weird/unexpected.

packages/react-ui-core/src/Gmap/GmapBase.js
tadjohnston: No chance because `initMap` doesn't get called until the component has mounted. The component won't mount until the google script has loaded.

packages/react-ui-core/src/Gmap/GmapBase.js
```diff
@@ -1,32 +1,107 @@
...
+        className={classnames(
+          theme.Gmap,
+          className,
+        )}
+        data-tid="google-map"
+        {...attributes}
       >
-        {children}
-      </GoogleMap>
+        <GmapContext.Provider value={{ map }}>
```
mpelzsherman: For my edification, can you explain why we are adding `GmapContext`? What does it buy us?

mpelzsherman: I did read the [docs](https://reactjs.org/docs/context.html), BTW, and they say "Don’t use context just to avoid passing props a few levels down. Stick to cases where the same data needs to be accessed in many components at multiple levels."

I take it the map components can get nested deep enough to warrant using context?

tadjohnston: > the same data needs to be accessed in many components at multiple levels. 

This is the key part right here.  Right now we may end up with a bunch of components that all need access to to `map`.  It also makes it so we don't have to clone every single child that the `GmapBase` will be a parent to.

packages/react-ui-core/src/Gmap/index.js
```diff
@@ -1,3 +1,5 @@
 export { default } from './Gmap'
 export { default as Gmap } from './Gmap'
 export { default as GmapSpinner } from './GmapSpinner'
+export { default as Marker } from './Marker'
+export { default as GmapInteraction } from './GmapInteraction'
```
orionstein:
```
export { GmapInteraction } from './GmapInteraction'
```
Since it's not export default. Either that, or make it a default export

tadjohnston: Good catch!

packages/react-ui-core/src/Gmap/GmapInteraction.js
```diff
@@ -0,0 +1,20 @@
...
+    this.registerMap = once(newMap => {
+      map = newMap
+    })
+
+    this.call = (f, args) => {
+      if (map[f] && typeof map[f] === 'function') {
+        return map[f](...args)
+      }
+    }
+    /* eslint-enable consistent-return */
```
mpelzsherman: Is it possible to return null or undefined here to make the linter happy, if nothing else?

tadjohnston: Done!

PR #229, opened 2018-04-03
chore(google map): fixing storybook marker

tadjohnston:
Accidently removed the marker position when I was testing.


tadjohnston:
Small change that doesn't affect any real code, so I'm marking this RTM.

PR #231, opened 2018-04-05
created Listing components for both Desktop and Mobile

okhanrt:
[Card](https://rentpath.leankit.com/card/646235681)


packages/react-ui-ag/src/Listings/SingleFamilyMobileListing.js
```diff
@@ -0,0 +1,68 @@
...
+    ratings: PropTypes.object,
+    propertyName: PropTypes.object,
+    ctaButtons: PropTypes.arrayOf(buttonPropTypes),
+    favoriteButton: buttonPropTypes,
+    lazyLoad: PropTypes.oneOfType([
+      PropTypes.object,
+      PropTypes.bool,
+    ]),
+    isActive: PropTypes.bool,
+    singleFamily: PropTypes.bool,
```
grenti: This boolean proptype doesn't seem to be used in this component. Can we remove it?

packages/react-ui-ag/src/Listings/SingleFamilyMobileListing.js
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  render() {
+    const { theme } = this.props
+
+    return (
+      <Listing
+        {...this.props}
+        className={theme.SingleFamilyMobileListing}
+        listingInfoComponents={this.getListingInfoComponents}
```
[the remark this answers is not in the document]
tadjohnston: Agree with @grenti. This is something we actually talked about a while ago.

packages/react-ui-ag/src/Listings/__tests__/SingleFamilyDesktopListing-test.js
```diff
@@ -0,0 +1,331 @@
...
+  photos: {
+    server: 'https://image.rent.com/',
+    dimensions: '280-120',
+  },
+  lazyLoad: false,
+  listingInfoComponents: () => {
+    const listing = baseListing
+    const { phone } = listing
+
+    // return (
```
nray621: Could you remove this commented out code?

nray621:
Aside from @grenti 's suggestions above, this looks good as far as I'm concerned.
Great job @okhanrt

packages/react-ui-ag/src/Listings/DesktopListing.js
```diff
@@ -0,0 +1,81 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+import themed from 'react-themed'
+import autobind from 'autobind-decorator'
+import { ListingComponents } from '@rentpath/react-ui-core'
+import Listing from './Listing'
+
+const buttonPropTypes = PropTypes.shape({
```
paudyalhari: can we move all custom prop type in separate file ?

packages/react-ui-ag/src/Listings/SingleFamilyMobileListing.js
okhanrt: Oops, forgot to remove that. Yes we can 👍

packages/react-ui-ag/src/Listings/__tests__/SingleFamilyDesktopListing-test.js
okhanrt: I forgot to remove that. Thanks for the catch! :)

packages/react-ui-ag/src/Listings/DesktopListing.js
tadjohnston: Why is that necessary?

packages/react-ui-ag/src/Listings/SingleFamilyMobileListing.js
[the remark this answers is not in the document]
okhanrt: @grenti I did think about combining SingleFamily listings with the `MobileListing` and `DesktopListing` components. It seemed like it would lead to many `singleFamily` conditional checks inside of `getListingInfoComponents` if new features were to be added. My approach of putting single family code in a different component was to make it less messy and somewhat maintainable in the future. However, since there are a few differences between the Single Family and regular listings and there is repeated code throughout, we could definitely combine it into `DesktopListing` and `MobileListing`.

Actually, the `ListingInfo` component is a good idea too. Or we can create two separate functions in Desktop/Mobile Listing called `getRegularListingInfo` and `getSingleFamilyListingInfo` that would appropriately get called depending on `singleFamily` prop.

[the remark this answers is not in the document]
grenti: @okhanrt I think the ListingInfo would be a good start so that it encapsulates the conditional and the components.

packages/react-ui-ag/src/Listings/DesktopListing.js
okhanrt: Oh since `buttonPropTypes` is the same in `DesktopListing` and `MobileListing`?

packages/react-ui-ag/src/Listings/SingleFamilyMobileListing.js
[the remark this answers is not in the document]
okhanrt: @grenti sounds like a plan. Thanks!

packages/react-ui-ag/src/Listings/DesktopListing.js
mpelzsherman: I'd say this is a "nice to have" refactor, but I'll still give you a +1 without it.

packages/react-ui-ag/src/Listings/Listing.js
```diff
@@ -0,0 +1,287 @@
...
+    if (isActive && onClick) {
+      onClick(this.props.listing, value)
+    } else if (!isActive && this.props.onClick) {
+      this.props.onClick(this.props.index, listing)
+    }
+
+    const shouldStopPropagation = isActive && event && event.stopPropagation
+
+    if (shouldStopPropagation) {
+      event.stopPropagation()
```
mpelzsherman: Will this break tagging?

packages/react-ui-ag/src/Listings/Listing.js
```diff
@@ -0,0 +1,287 @@
...
+    }
+  }
+
+  @autobind
+  handlePhotoCarouselSlide(index) { // eslint-disable-line no-unused-vars
```
mpelzsherman: Can you just omit `index` here?

okhanrt: Oh right, it's not being used. Sure, it doesn't break the feature.

okhanrt: I just remembered why I was passing index parameter into this function. It was to implement photo count on the carousel but we later found out we can pass `showIndex` prop into `PhotoCarousel` to display photo count. So yes, we can definitely omit `index` here.

packages/react-ui-ag/src/Listings/Listing.js
okhanrt: Good question, I'm not sure. I copied this function from `MobileMapListing` and this component is currently being used on mobile.

okhanrt: @tadjohnston what do you think?

stories/ag/Listings/DesktopListing.js
```diff
@@ -0,0 +1,98 @@
...
+  },
+  phone: '(404) 770-5555',
+}
+
+const props = {
+  listing: baseListing,
+  onClick: () => action('click')('listing cell click'),
+  favoriteButton: {
+    onClick: (listing, value) => action('click')('favorite toggled to ', value),
+    children: '♥',
```
mpelzsherman: Seeing this always bugs me a little. I know we're going to replace it with an SVG in ag.js, but why can't `react-ui-ag` provide that out of the box?

mpelzsherman:
Seeing some odd behavior in storybook on the MobileListing component. The left scroll click area extends far below the carousel and into the listing info section.

http://recordit.co/Q5TafXv0yj

packages/react-ui-ag/src/Listings/DesktopListing.js
tadjohnston: Yeah, I don't think this is truly necessary and once we move to TypeScript it'll be irrelevant as they will be in the definition files, but I do like the thought process @paudyalhari 👍

packages/react-ui-ag/src/Listings/Listing.js
tadjohnston: It all depends on how the tag is fired.  If it's the normal way, no, because the `stopPropagation` here is related to the synthetic React event while the tagging event is completely outside of React.

packages/react-ui-ag/src/Listings/Listing.js
```diff
@@ -0,0 +1,287 @@
...
+    if (lazyLoad && typeof lazyLoad === 'boolean') {
+      lazyLoad = typeof lazyLoad === 'boolean'
+        ? REACT_LAZYLOAD_PROP_DEFAULTS
+        : lazyLoad
+    }
+
+    if (!isActive) {
+      const { server, dimensions } = photos
+      const firstPhoto = this.listingPhotos[0] || {}
+      listing.photo = { url: `${server}${firstPhoto.path}${dimensions}` }
```
paudyalhari: can we use allPhotos function from photos utility(apps/shared/Listings/photos.js)? if we will use that utility we do not need to have photos object listing photo url will be taken care by photos utility like we did in DesktopMapPinListing.js.

stories/ag/Listings/DesktopListing.js
mpelzsherman: Not a blocker for the PR but I would like to see this addressed some day?

packages/react-ui-ag/src/Listings/Listing.js
mpelzsherman: I'd say this can wait for a future refactor. @tadjohnston ?

stories/ag/Listings/DesktopListing.js
tadjohnston: I'm not sure why it bugs you.  This was an architecture decision.  How are you suggesting we add the svg in react-ui-ag, through a sprite, directly?   If the icon changes for some reason we now have to edit it here, publish and then bump the version which is seems somewhat superfluous.  We don't do this is this is exact reason. Any visual styling is generally done outside of the scope of react-ui. We can talk about this if you'd like if you have ideas on how we can separate these out and include icons and styling.

packages/react-ui-ag/src/Listings/Listing.js
tadjohnston: Yeah, I like the idea @paudyalhari.  We can fix this later as we should probably rethink how we want to do photos in react-ui.

paudyalhari:
👍

PR #233, opened 2018-04-06
refactor(google maps): refactoring setup

tadjohnston:

- tweaking gmap to now pass the map to the children instead of context.
- added HOC for the google map script
- added multiple marker support with geojson data
- refactored map to better suit needs


packages/react-ui-core/src/Gmap/Gmap.js
```diff
@@ -1,86 +1,162 @@
...
+
+const EVENTS = {
+  onBoundsChanged: 'bounds_changed',
+  onCenterChanged: 'center_changed',
+  onClick: 'click',
+  onDoubleClick: 'dblclick',
+  onDrag: 'drag',
+  onDragEnd: 'dragend',
+  onDragStart: 'dragstart',
+  onHeadingChnaged: 'heading_changed',
```
ninomichaels: `onHeadingChnaged` misspelled.

packages/react-ui-ag/src/Gmap/Gmap.js
```diff
@@ -0,0 +1,24 @@
...
+  render() {
+    const {
+      apiKey,
+      ...rest
+    } = this.props
+
+    return (
+      <GoogleMap
+        apiKey={apiKey}
+        spinner={<GmapSpinner color="#ff0000" loading closeDelay={1000} />}
```
ninomichaels: don't need to make `closeDelay` configurable?

packages/react-ui-core/src/Gmap/Gmap.js
```diff
@@ -1,86 +1,162 @@
...
+  onIdle: 'idle',
+  onMaptypeIdChanged: 'maptypeid_changed',
+  onMouseMove: 'mousemove',
+  onMouseOut: 'mouseout',
+  onMouseOver: 'mouseover',
+  onProjectionChanged: 'projection_changed',
+  onResize: 'resize',
+  onRightclick: 'rightclick',
```
ninomichaels: `onRightClick` instead of `onRightclick`?

packages/react-ui-core/src/Gmap/Markers.js
```diff
@@ -0,0 +1,69 @@
+import { Component } from 'react'
+import PropTypes from 'prop-types'
+import { setupMarker, removeMarker } from './utils/markerHelpers'
+
+const MARKER = ({ properties, geometry }) => ({
+  id: properties.id,
+  position: {
+    lat: geometry.coordinates[1],
```
ninomichaels: is this order guaranteed? also seems a bit weird, as`lat` always come before `lng` normally

packages/react-ui-core/package.json
```diff
@@ -31,6 +31,7 @@
     "babel-core": "6.26.0",
     "babel-eslint": "^8.2.2",
     "babel-jest": "^22.4.3",
+    "babel-plugin-dynamic-import-node": "^1.2.0",
```
orionstein: Is this actually working anywhere? I think we only removed an import.then, I'm not sure where it's happening

packages/react-ui-core/src/Gmap/utils/mapEventHelpers.js
```diff
@@ -0,0 +1,13 @@
+export const handleEvent = (instance, name, props) =>
+  event => props[name](event, props, instance)
+
+export const setupEvents = (instance, events, props) => {
+  const eventNames = Object.keys(events)
+
+  eventNames.forEach(name => {
+    if (props[name] && typeof props[name] === 'function') {
+      const event = handleEvent(instance, name, props)
+      window.google.maps.event.addListener(instance, events[name], event)
```
ninomichaels: Is this assuming too much? (that `window.google.maps...` will be available?)

packages/react-ui-core/src/Gmap/Gmap.js
```diff
@@ -1,86 +1,162 @@
...
-    await import(/* webpackChunkName: "scriptjs" */ 'scriptjs').then(scriptjs => {
-      scriptjs(url, this.scriptLoaded)
+    this.map = new window.google.maps.Map(this.googleMap.current, {
+      zoom,
+      clickableIcons,
+      center: center || DEFAULT_CENTER,
+      ...mapControls,
     })
+
+    this.setupEvents()
```
orionstein: Does it make sense to call mapEventHelpers setupEvents() here since that logic exists?

packages/react-ui-core/package.json
```diff
@@ -54,15 +55,16 @@
     "form-serialize": "0.7.2",
     "mapbox-gl": "0.42.2",
     "prop-types": "^15.6.1",
+    "react": "16.3.1",
```
tcwilson: there a reason you added this here? I just had to fix a bug in hive related to this sort of thing. `dependencies` are really for private packages of the library. Since `react` is used by the app (rent-js/ag.js) and other packages, then all versions have to be in agreement. Which is what `peerDependencies` is for, for library packages.

Let's say rent-js installed 16.3.2, and this is locked to 16.3.1. Then you end up with:
```
node_modules/react  <-- 16.3.2
node_modules/@rentpath/react-ui-core/node_modules/react  <-- 16.3.1
```
Without aliases, webpack is going to bundle both of these versions which can cause JS errors.

(interesting note about this... `yarn link` also is problematic in similar manner if you `yarn install` inside a library package and then link that package to an app... npm, I swear 🤦‍♂️  )

packages/react-ui-ag/src/Gmap/Gmap.js
orionstein: I think the idea is that this is the configurable part - if we wanted to update the GmapSpinner for ag or rent, we would just update it here in there respective Gmap files - this would also apply to adding styles/classes etc, I believe

packages/react-ui-core/package.json
orionstein: We could add react to the resolutions section, perhaps

```"resolutions": {
    "koa": "2.3.0",
    "react": "16.3.2",
}```

File: packages/react-ui-core/src/Gmap/utils/mapEventHelpers.js (hunk 6 above)
Comment (orionstein) [reply packages/react-ui-core/src/Gmap/utils/mapEventHelpers.js]: I mean, if we don't have google maps, we have a problem.

However, yeh, might just want to make an isGoogleMapsAvailable checker or something so we don't throw errors

File: packages/react-ui-core/src/Gmap/Gmap.js (hunk 3 above)
Comment (tadjohnston) [reply packages/react-ui-core/src/Gmap/Gmap.js]: Nice catch!

File: packages/react-ui-core/src/Gmap/Gmap.js [hunk 9]
```diff
@@ -1,86 +1,162 @@
...
   }
 
   static defaultProps = {
-    libraries: [],
+    theme: {},
+    zoom: DEFAULT_ZOOM,
+    clickableIcons: false,
+    mapControls: {
+      fullscreenControl: false,
+      streetViewControl: false,
```
tcwilson: what about `mapTypeControl` and `zoomControl` defaults?

packages/react-ui-core/src/Gmap/Markers.js
tadjohnston: Yeah I dunno, I should double check the geojson standard. I took this directly from ag.js so who knows if it's right lol.

packages/react-ui-core/src/Gmap/utils/mapEventHelpers.js
tadjohnston: Yeah, we can add a checker, but yeah, the way these work with the markers and such is that the api has to be loaded and the map has to be initialized before any of the children are mounted.

packages/react-ui-core/src/Gmap/Markers.js
```diff
@@ -0,0 +1,69 @@
...
+
+    this.markers[id] = setupMarker(map, marker)
+    return this.markers[id]
+  }
+
+  render() {
+    const { map, geojson } = this.props
+
+    if (map) {
+      geojson.features.map(feature => this.renderMarker(map, feature))
```
tcwilson: if I'm understanding this correctly, this just keeps appending to the visible markers when `geojson` updates? Rather than clear out existing markers?

tadjohnston: We should probably clear first... I'll add that in.

packages/react-ui-core/package.json
tcwilson: Well, certainly not by using two versions of react at once 🤣 

The `peerDependencies` already has `"16.3.1"`, which should cause a warning on rent-js if that library version is not available (though it will be, since I put in a PR for it). I'm not sure why we locked those down (really, `"^16.3.1"` will give you versions >= 16.3.1 and < 17.0.0).

We really need to clean up peer dependencies on rent-js and start paying attention to the warnings. Letting the app  specify the version on react/react-dom and other "global" sort of packages is, IMO, the [right way to go](https://stackoverflow.com/a/34645112).

tadjohnston: We locked them down specifically for the reason this change addresses.  I see your point though, what do you think about adding `react-dom` as a peer as well?

packages/react-ui-core/src/Gmap/Gmap.js
tadjohnston: We can add them, I didn't want to be overly aggressive when adding these as defaults.

packages/react-ui-core/src/Gmap/Gmap.js
tadjohnston: Nope, good catch. It should be using the helper for this.

packages/react-ui-core/src/Gmap/Markers.js
ninomichaels: looks like Mapbox and GeoJSON prefer `long, lat`, but Google Maps API prefers `lat, long`, so, yayyyy

tadjohnston: The good thing is you can still override this if you wanted 😄

tadjohnston: From the [RFC](https://tools.ietf.org/html/rfc7946)
```
9.  Mapping 'geo' URIs

   'geo' URIs [RFC5870] identify geographic locations and precise (not
   uncertain) locations can be mapped to GeoJSON Geometry objects.

   For this section, as in [RFC5870], "lat", "lon", "alt", and "unc" are
   placeholders for 'geo' URI latitude, longitude, altitude, and
   uncertainty values, respectively.

   A 'geo' URI with two coordinates and an uncertainty ('u') parameter
   that is absent or zero, and a GeoJSON Point geometry may be mapped to
   each other.  A GeoJSON Point is always converted to a 'geo' URI that
   has no uncertainty parameter.

   'geo' URI:

   geo:lat,lon

   GeoJSON:

   {"type": "Point", "coordinates": [lon, lat]}

   The mapping between 'geo' URIs and GeoJSON Points that specify
   elevation is shown below.

   'geo' URI:

   geo:lat,lon,alt

   GeoJSON:

   {"type": "Point", "coordinates": [lon, lat, alt]}

   GeoJSON has no concept of uncertainty; imprecise or uncertain 'geo'
   URIs thus cannot be mapped to GeoJSON geometries.
```

tadjohnston: Seriously? So they don't even follow the geojson standard?

packages/react-ui-ag/src/Gmap/Markers.js
```diff
@@ -0,0 +1,40 @@
...
+      },
+      ...this.props.marker(),
+    })
+  }
+
+  render() {
+    return (
+      <Gmarkers
+        marker={this.marker}
+        {...this.props}
```
tcwilson: does the order of these 2 lines need switched? You have a `marker` prop, which I think will override the `marker={this.marker}`, right?

tadjohnston: Good catch. It should be destructured from this.props instead 👍

packages/react-ui-core/src/Gmap/Markers.js
```diff
@@ -0,0 +1,83 @@
...
+    this.markers = EMPTY_MARKERS
+  }
+
+  marker(feature) {
+    const props = this.props.marker(feature)
+    const defaults = MARKER(feature)
+
+    if (typeof props === 'object') {
+      return {
+        ...defaults,
```
lfender6445: not a big deal but since you only use defaults if props is an object, you could do

```
{
   ...(MARKER(feature)),
   ...props,
}
```

packages/react-ui-core/src/Gmap/utils/markerHelpers.js
```diff
@@ -0,0 +1,36 @@
...
+    window.google.maps.event.clearInstanceListeners(marker)
+    marker.setMap(null)
+  }
+}
+
+export const setupMarker = (map, props) => {
+  if (map) {
+    const propsWithoutEvents = omitBy(props, (key, val) => EVENTS[val])
+    const marker = new window.google.maps.Marker({
+      ...propsWithoutEvents,
```
lfender6445: for my curiosity, why does a marker not get events?

packages/react-ui-core/src/Gmap/Markers.js
tadjohnston: That would probably be cleaner 👍

packages/react-ui-core/src/Gmap/utils/markerHelpers.js
tadjohnston: It does, but they have to be added after the marker has been instantiated. We could probably leave it "as is" and pass the events through as well as I don't think it would hurt anything.

packages/react-ui-core/src/Gmap/withGoogleScript.js
```diff
@@ -0,0 +1,92 @@
...
+import PropTypes from 'prop-types'
+import autobind from 'autobind-decorator'
+import scriptjs from 'scriptjs'
+import { getDisplayName } from 'recompose'
+import { parseArgs } from '@rentpath/react-ui-utils'
+
+const API_BASE_URL = 'https://maps.googleapis.com/maps/api/js'
+
+export default function(BaseComponent) {
+  const factory = React.createFactory(BaseComponent)
```
lfender6445: according to docs this method is discouraged

would something like this work? 

```
const withGoogleScript = Wrapped => wrappedProps => {
  <Wrapped ...wrappedProps><Container /></Wrapped>
}
```

packages/react-ui-core/src/Gmap/withGoogleScript.js
```diff
@@ -0,0 +1,92 @@
+import React, { PureComponent } from 'react'
+import PropTypes from 'prop-types'
+import autobind from 'autobind-decorator'
+import scriptjs from 'scriptjs'
+import { getDisplayName } from 'recompose'
```
tcwilson: how much does `recompose`+dependencies add to the bundle? If we're just using `getDisplayName`, you might just want to copy that function over and use it, since it's [fairly small](https://github.com/acdlite/recompose/blob/master/src/packages/recompose/getDisplayName.js).

You might also want to consider doing:
`import getDisplayName from 'recompose/getDisplayName`
and compare bundle sizes, see if that cuts it down.

packages/react-ui-core/src/Gmap/utils/markerHelpers.js
lfender6445: oh no this is good. just wondering

packages/react-ui-core/src/Gmap/withGoogleScript.js
tadjohnston: It's discouraged as it's legacy. `React.createElement` should be used instead which is all jsx is under the hood anyway.

packages/react-ui-core/src/Gmap/utils/markerHelpers.js
tadjohnston: Yeah, I wish we could just pass them in, I don't understand why you can't. 🤷‍♂️

packages/react-ui-core/src/Gmap/withGoogleScript.js
tadjohnston: I'm not sure this is better:
```

packages/react-ui-core/src/Gmap/Markers.js
lfender6445: oh, forget what i said. i see now defaults is used if the first condition fails

lfender6445:
Lookin good 👉😎👉

tcwilson:
I think I'm good with this. 👍

PR #236, opened 2018-04-09
fix(google map): ssr map loading broken

tadjohnston:

`scriptjs` only works client side and actually breaks SSR, so let's just do good ole fashioned adding the script to the head.

### DIFF
--- a/packages/react-ui-core/src/Gmap/withGoogleScript.js
+++ b/packages/react-ui-core/src/Gmap/withGoogleScript.js
@@ -1,7 +1,6 @@
 import React, { PureComponent } from 'react'
 import PropTypes from 'prop-types'
 import autobind from 'autobind-decorator'
-import scriptjs from 'scriptjs'
 import { parseArgs } from '@rentpath/react-ui-utils'
 import getDisplayName from './utils/getDisplayName'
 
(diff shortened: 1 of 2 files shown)


packages/react-ui-core/src/Gmap/withGoogleScript.js
```diff
@@ -65,8 +64,14 @@ export default function(BaseComponent) {
...
       const { apiKey, libraries, version } = this.props
-      const url = `${API_BASE_URL}?key=${apiKey}&version=${version}&libraries=${libraries.join()}`
-      scriptjs(url, this.scriptLoaded)
+      const script = document.createElement('script')
+
+      window.google_map_initialize = this.scriptLoaded
+
+      script.async = true
+      script.defer = true
+      script.src = `${API_BASE_URL}?key=${apiKey}&version=${version}&libraries=${libraries.join()}&callback=google_map_initialize`
```
orionstein: Should we add a script.onerror in case the script doesn't load correctly? We could add a window flag if the map errors out we could check, potentially, not sure where else we would put it - since, google_map_initialize would never run, we would have no way to handle that. I guess we wouldn't have a plan of what to do with it if it did fail, but we would have the option to do so (would be something to bring up with team prod)

tadjohnston: Done!

tcwilson:
👍

lfender6445:
👍

orionstein:
👍

PR #240, opened 2018-04-13
Switching from react-picker to simple datepicker using date-fns

orionstein:
We didn't want to have to include moment since our bundling is questionable atm, so we worked on swapping out react-datepicker with something that uses something modular, like date-fns.


packages/react-ui-core/src/DateInput/DateInput.js
```diff
@@ -1,24 +1,33 @@
...
-   return (
-     <DatePicker {...this.props} />
-   )
- }
+    return (
+      <DatePicker
+        handleDateChange={onChange}
+        value={selected}
+        datepickerName={name}
+        datepickerClassName="DatePicker"
```
barelyfitz: This component is `DateInput` - should the classname also be `DateInput` for consistency?

orionstein: Actually, this is the classname for the picker itself - that goes to the child, since it's datepickerClassName, rather than className. I think it makes sense like this, but we could always change it

packages/react-ui-core/package.json
```diff
@@ -55,6 +55,7 @@
     "form-serialize": "0.7.2",
     "mapbox-gl": "0.42.2",
     "prop-types": "^15.6.1",
+    "react-basic-datepicker": "rentpath/react-basic-datepicker",
     "react-datepicker": "^1.3.0",
```
tadjohnston: We should probably get rid of this.

tadjohnston:
Please also make sure you use `git cz` for commits 👍

packages/react-ui-core/src/DateInput/DateInput.js
```diff
@@ -1,24 +1,33 @@
...
+import DatePicker from 'react-basic-datepicker'
 
 export default class DateInput extends Component {
+  static propTypes = {
+    selected: PropTypes.object,
+    onChange: PropTypes.func,
+    maxDate: PropTypes.object,
+    minDate: PropTypes.object,
+    placeholderText: PropTypes.string,
+    className: PropTypes.string,
```
barelyfitz: `className` prop doesn't seem to be used at all (not passed through to `<DatePicker>`

packages/react-ui-core/src/DateInput/DateInput.js
```diff
@@ -1,24 +1,33 @@
 import React, { Component } from 'react'
 import PropTypes from 'prop-types'
-import DatePicker from 'react-datepicker'
+import DatePicker from 'react-basic-datepicker'
 
 export default class DateInput extends Component {
+  static propTypes = {
+    selected: PropTypes.object,
+    onChange: PropTypes.func,
+    maxDate: PropTypes.object,
```
barelyfitz: The props `maxDate` and `minDate` don't seem to be modified at all in this component (no default value, not passed explicitly to `<DatePicker>` except through `{...rest}` like any other unknown props) so should they just be omitted from propTypes?

It's probably nice to document which props we expect to normally support for `DateInput` but that could be done with some comments.

packages/react-ui-core/src/DateInput/__tests__/DateInput-test.js
```diff
@@ -5,16 +5,16 @@ import DateInput from '../DateInput'
...
-    const wrapper = shallow(<DateInput id="foo" />)
-    expect(wrapper.find('DatePicker').prop('id')).toBe('foo')
+    const wrapper = shallow(<DateInput datepickerId="foo" />)
+    expect(wrapper.html().includes('id="foo"')).toEqual(true)
   })
 
   it('applies a custom className', () => {
     const wrapper = shallow(<DateInput className="DatePicker" />)
-    expect(wrapper.find('DatePicker').prop('className')).toBe('DatePicker')
+    expect(wrapper.html().includes('class="DatePicker"')).toEqual(true)
```
barelyfitz: It looks like the code now hard-codes `DatePicker` classname rather than using the `className` prop, so though this test might be passing it doesn't look like a valid test. Remove the test if we're not supporting custom `className` prop, or modify `DateInput` so it applies the `className` prop.

orionstein: I need to update that, that was supposed to use className as the main option, will fix

packages/react-ui-core/src/DateInput/DateInput.js
tadjohnston: Is there a way to pass in whatever you want? Like could we theme this component?

packages/react-ui-core/src/DateInput/DateInput.js
```diff
@@ -1,24 +1,30 @@
...
-   return (
-     <DatePicker {...this.props} />
-   )
- }
+    return (
+      <DatePicker
+        handleDateChange={onChange}
+        value={selected}
+        datepickerName={name}
+        datepickerClassName={className || 'DatePicker'}
```
barelyfitz: Could this just be a defaultProp instead?

packages/react-ui-core/src/DateInput/__tests__/DateInput-test.js
```diff
@@ -1,20 +1,20 @@
...
 
 describe('DateInput', () => {
   it('renders a DateInput', () => {
-    const wrapper = shallow(<DateInput />)
-    expect(wrapper.find('DatePicker')).toHaveLength(1)
+    const wrapper = mount(<DateInput />)
+    expect(wrapper.find('Datepicker')).toHaveLength(1)
   })
 
-  it('passes through props', () => {
```
barelyfitz: Can you add this test back in to test if passing in `unknownProp=unknownValue` gets passed through to `DatePicker`?

orionstein: I wouldn't expect it to, since the plugin itself doesn't spread props - they would get lost there

barelyfitz: In `render()` method you are passing `{...rest}` to the `DatePicker` component. So for example, if I wanted to set a DatePicker prop I should be able to pass it through this component.

orionstein: I guess I should remove rest, I suppose, and only pass through all the props that the  plugin expects

PR #242, opened 2018-04-13
refactor(react-ui-core/Gmap): fixing call for deep functions, adding …

orionstein:
…Marker Interaction

Fixing call on MapInteractions for allowing use of call('data.foo', [...args])


tcwilson:
👍

PR #243, opened 2018-04-13
fix(ListingCarousel): increases debounce time in ListingCarousel

yasabere:

Increases the debounce time in ListingCarousel


bethstevic:
Looks good to me 👍

nray621:
I was going to ask if the debounce time was increased too much, but then I realized if you're trying to tap twice in 1.2 seconds you need to chill

nray621:
LGTM 👍

tadjohnston:
Please provide an explanation as to why the increase and a corresponding ticket.

PR #244, opened 2018-04-16
fix(react-ui-ag Listing component): listing component is not rerender…

okhanrt:
…ing on isFavorited prop change


paudyalhari:
👍

PR #245, opened 2018-04-16
Adds a Layer component for showing geojson

paulpostel:
[Card](https://rentpath.leankit.com/card/646884884)

### DIFF
--- a/packages/react-ui-core/src/Gmap/Layer.js
+++ b/packages/react-ui-core/src/Gmap/Layer.js
@@ -1,0 +1,42 @@
+import { PureComponent } from 'react'
+import PropTypes from 'prop-types'
+
+export default class Layer extends PureComponent {
+  static propTypes = {
+    map: PropTypes.object,
+    geojson: PropTypes.object.isRequired,
+  }
+
+  constructor(props) {
+    super(props)
+    this.features = new Set()
+  }
+
+  componentWillUnmount() {
+    this.removeFeatures()
+  }
+
+  removeFeatures() {
+    const { map } = this.props
+
+    if (map) {
+      this.features.forEach(feature => {
+        map.data.remove(feature)
+      })
+    }
+    this.features.clear()
+  }
+
+  render() {
+    const { map, geojson } = this.props
+
+    this.removeFeatures()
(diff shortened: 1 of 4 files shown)


packages/react-ui-core/src/Gmap/Layer.js
```diff
@@ -0,0 +1,58 @@
...
+  componentWillUnmount() {
+    this.removeFeatures()
+  }
+
+  removeFeatures() {
+    const { map } = this.props
+
+    this.features.forEach(feature => {
+      if (map) map.data.remove(feature)
+      this.features.delete(feature)
```
tcwilson: It's probably slightly more efficient to do `this.features.clear()` after the `forEach`. But if this set only has a few hundred items it probably doesn't matter much.

tcwilson:
👍

packages/react-ui-core/src/Gmap/Layer.js
tadjohnston: If it's ever possible it would have more than that, I'd prefer the clear

tadjohnston:
@paulpostel here is mine that was gonna create for ag.js map markers: https://gist.github.com/tadjohnston/1c250234245e343a7b333302ffc19076

If you see anything in there you might want to use, feel free to add it. The hope was that the GeojsonMarkers I was creating would eventually make it into react-ui-core.

paulpostel:
@tcwilson pointed out that the `data.setStyle` function would accept a function argument that receives the feature-to-be-styled.  That means I wouldn't have to `data.overrideStyle` for each feature in the `Layer`.  since the styling function is one-per-map, what do you think about adding it as an optional prop to `Gmap`?  I guess it could also be set thru a separate `GmapData` component or something.

packages/react-ui-core/src/Gmap/Layer.js
paulpostel: agreed, `clear()` is better, i'll update

paulpostel:
@tadjohnston, @tcwilson I added the styling function to `Gmap` as described above.  I think it's a better way to go.  Please take a another look at this.

PR #246, opened 2018-04-16
fix(react-ui-core): fixing deps and babelrc for fresh installs

orionstein:

Some dependency changes broke jest on fresh installs. This installs the missing dependency and
plugin for babel to transpile and run our tests correctly


lfender6445:
I've confirmed the test runner was broken on fresh install of a machine without these changes - im guessing babel-stage-0 was a dep of a dep once upon a time and was removed at another point? 

Do we want to consider locking down babel/jest related dependencies?

tadjohnston (dismissed):

PR #251, opened 2018-04-21
fix(google maps): swap lat/lng to match geojson spec; efficiency tweaks

tcwilson:

- Fixes constant re-rendering of `Markers` component. 
- Marker lng/lat order fixed to match GeoJSON spec.
- Various other slight efficiency tweaks.


tcwilson:
cc @paulpostel

PR #252, opened 2018-04-25
feat(google maps): add Google Maps InfoWindow component

tcwilson:


stories/core/Marker.js
```diff
@@ -1,24 +0,0 @@
-import React, { PureComponent } from 'react'
-import { action } from '@storybook/addon-actions'
-import { Marker as Gmarker } from 'react-ui-core/src'
-
-export default class Marker extends PureComponent {
```
tcwilson: I couldn't find where this was used, and it looks like it was replaced in the storybook at some point. So I removed it.

PR #254, opened 2018-04-25
feat(google maps): add Google Maps InfoWindow and OverlayView popups

tcwilson:


packages/react-ui-core/src/Gmap/InfoWindow.js
```diff
@@ -0,0 +1,79 @@
...
+
+  constructor(props) {
+    super(props)
+
+    this.initInfoWindow()
+  }
+
+  componentDidMount() {
+    this.infowindow.setContent(this.container)
+    this.openWindow()
```
tadjohnston: Do we really want to open this on mount or should do the same exact check for `anchor` we do on update?

packages/react-ui-core/src/Gmap/__tests__/GmapInteraction-test.js
```diff
@@ -57,7 +57,7 @@ describe('MarkerInteraction', () => {
     const result = GmapInteraction.MarkerInteraction.setupMarker({})
 
     expect(markerSpy).toHaveBeenCalled()
-    expect(result).toEqual({})
+    expect(typeof result.getPosition).toEqual('function')
```
tadjohnston: 👍

packages/react-ui-core/test/shim.js
```diff
@@ -22,9 +22,28 @@ global.google = {
       getSouthWest: () => sw,
       getNorthEast: () => ne,
     }),
-    OverlayView: () => ({}),
-    InfoWindow: () => ({}),
-    Marker: () => ({}),
+    OverlayView: () => ({
```
tadjohnston: Nice additions!

packages/react-ui-core/src/Gmap/InfoWindow.js
tadjohnston: Good point! I'm good with it then 😄

packages/react-ui-core/src/Gmap/OverlayView.js
```diff
@@ -0,0 +1,110 @@
...
+
+  static defaultProps = {
+    className: '',
+    theme: {},
+  }
+
+  constructor(props) {
+    super(props)
+
+    this.initOverlayView()
```
tadjohnston: This will probably fail on SSR since `window` isn't available in `initOverlayView`.

tcwilson: Good point. I was thinking that because it's used as a child of `Gmap`, it will be guaranteed to render only in the browser. But one of the benefits of passing `map` as a prop rather than context is that `OverlayView` doesn't necessarily need to be a child of `Gmap` now.

Anyway, I moved that call to `componentDidMount` so this should be good now.

tadjohnston: Yeah, I was thinking of it being a child if it made a difference or not. I guess better to be safe than sorry 🤷‍♂️

PR #255, opened 2018-04-25
Fix invalid use of `String.prototype.replace()`

michaelstalker:
The method takes two arguments, but we were passing in one. A
replacement would result in "undefined" being injected into the output
string.


bethstevic:
Looks good to me 👍

PR #257, opened 2018-04-26
Added placeholder for carousel image tagging

paudyalhari:
[Card](https://rentpath.leankit.com/card/610515393)

* Added onImageClick function to handle only image click

### DIFF
--- a/packages/react-ui-core/src/Carousel/Carousel.js
+++ b/packages/react-ui-core/src/Carousel/Carousel.js
@@ -28,6 +28,7 @@ export default class Carousel extends Component {
     lazyLoad: PropTypes.bool,
     selectedIndex: PropTypes.number,
     onSlide: PropTypes.func,
+    onImageClick: PropTypes.func,
     showNav: PropTypes.bool,
     navigation: PropTypes.shape({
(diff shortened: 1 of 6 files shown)


tadjohnston:
Why can't you just use `onClick` which would already work?

packages/react-ui-core/src/Carousel/Carousel.js
```diff
@@ -70,6 +71,12 @@ export default class Carousel extends Component {
     }
   }
 
+  @autobind
+  onImageClick() {
+    this.props.onImageClick()
+    console.log('Inside carousel')
+  }
+
```
bethstevic: This won't be needed. It can be done in ag.js

bethstevic: I am about to push a PR to take care of that

paudyalhari:
@tadjohnston actually the existing onClick is not only for image, its handling the whole card onClick which is triggering the 'image' tag while clicking on PropertyName too. but I guess I found the another solution.

paudyalhari:
Closing this pr since Beth already created another PR for propertyName

PR #260, opened 2018-04-26
feat(google maps): add onMarkersReady event to access Google Markers …

tcwilson:
…by listing id


paulpostel:
👍 @tcwilson, I have https://github.com/rentpath/react-ui/pull/262 which has other Gmap changes i need.  Can we coordinate the publish step for these two?

PR #263, opened 2018-05-01
fix(ListingCell): Allow rerender on classname change

yasabere:

-Allow rerender on classname change


nray621:
LGTM 👍

PR #268, opened 2018-05-02
simplify ratings bar implementation

ninomichaels:
before:
screen shot 2018-05-02 at 11 54 11 am

after:
screen shot 2018-05-02 at 11 52 21 am


tadjohnston:
I think it's been annoying everyone @paulpostel 😄

stories/theme/core/RatingBar.css
```diff
@@ -1,18 +1,44 @@
...
+  height: 20px;
+  width: 100px;
+  background-repeat: repeat-x;
 }
 
-.RatingBar_Item svg {
-  height: 20px;
-  width: 20px;
+.RatingBar_Star {
+  background-image: GoldStar);
```
barelyfitz: Assume the weird syntax here is working around some kind of problem - is there a better solution? I don't like the idea of having such a malformed rule.

stories/theme/core/RatingBar.css
```diff
@@ -1,18 +1,44 @@
...
+.RatingBar_StarBackground {
+  background-image: GrayStar);
+}
+
+.RatingBar_StarMax10 {
+  background-image: GoldStar);
+}
+
+.RatingBar_StarMax10Background {
+  width: 200px;
```
barelyfitz: In the story, the Max10 example is showing only five stars.

packages/react-ui-core/src/Ratings/__tests__/__snapshots__/RatingBar-test.js.snap
```diff
@@ -0,0 +1,9 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`RatingBar ratings fills 1 of 5 rating items 1`] = `"<div class=\\"\\"><div class=\\"\\"><div class=\\"\\" style=\\"width:20%\\"></div></div></div>"`;
```
barelyfitz: These snapshots are not getting the expected classnames, perhaps need to mock a theme to be passed into the component so the classnames get added.

Need to add coverage for a developer specifying a different classname either using className or RatingIcon props.

packages/react-ui-core/src/Ratings/__tests__/RatingBar-test.js
```diff
@@ -1,72 +1,29 @@
...
-      const wrapper = mount(<RatingBar
-        uniqueId="foo"
-        maxScore={1}
-        theme={theme}
-        fillColor="red"
-      />)
-      expect(wrapper.find('.RatingBar_Item-red').length).toBe(0)
-    })
+    const wrapper = shallow(<RatingBar maxScore={5} />)
+    expect(wrapper.html()).toMatchSnapshot()
```
barelyfitz: If only testing the rendered HTML of the component, consider using enzyme's `render` instead of `shallow` (but I'm actually not sure what the prefered method is).

stories/theme/core/RatingBar.css
```diff
@@ -1,18 +1,44 @@
...
 }
 
-.RatingBar_Item {
-  display: inline-block;
+.RatingBar_Background {
+  overflow: hidden;
+  max-width: 100px;
+  height: 20px;
+  width: 100px;
+  background-repeat: repeat-x;
```
barelyfitz: Since we are using `background-repeat` to fill bar with five stars, there are a lot of assuptions going on here:
* The width of the image is 20px
* We set the width to 100px (5 times 20px)

It would be good at least add some comments here:

```
width: 100px; /* Star width (20px) times 5 stars */
```

packages/react-ui-core/src/Ratings/RatingBar.js
```diff
@@ -2,60 +2,65 @@ import React, { PureComponent } from 'react'
...
   static propTypes = {
-    uniqueId: PropTypes.string.isRequired,
     theme: PropTypes.object,
-    fillColor: PropTypes.string,
     className: PropTypes.string,
     maxScore: PropTypes.number,
     score: PropTypes.number,
     label: PropTypes.string,
-    RatingItem: PropTypes.func,
+    RatingIcon: PropTypes.string,
```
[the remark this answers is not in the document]
ninomichaels: good notes @barelyfitz  thanks!

barelyfitz:
Theme for AG / ListingsCarousel needs to be updated (rating bar on the ListingsCarousel story is not displaying as expected).

barelyfitz:
Story "react-ui-ag / Listings / Desktop Listing" not displaying rating bar.

ninomichaels:
Storybooks for AG are fixed.

stories/core/RatingBar.js
```diff
@@ -2,6 +2,14 @@ import React from 'react'
 import { RatingBar } from 'react-ui-core/src'
 import theme from '../theme/core/RatingBar.css'
 
+const handleRatingClick = ({ pageX, currentTarget }) => {
+  const getBoundingClientRect = currentTarget.getBoundingClientRect()
+  const starValue =
+    Math.ceil(((pageX - getBoundingClientRect.x) / getBoundingClientRect.width) * 5)
+  // eslint-disable-next-line no-alert
+  window.alert(starValue)
```
barelyfitz: Rather than an alert, in the react-ui stories you can do something like this:
```
import { action } from '@storybook/addon-actions'
action(`Clicked star value`)(starValue)
```

packages/react-ui-core/src/Ratings/RatingBar.js
```diff
@@ -2,60 +2,54 @@ import React, { PureComponent } from 'react'
...
-        width={`${this.fillWidth(index)}%`}
-        {...props}
-      />
-    ))
+    const calcScore = (score / maxScore)
+    const scorePercent = (calcScore > 1 ? 1 : calcScore) * 100
+
+    return (
+      <div
+        className={theme.RatingBar_Background} {...props}
```
barelyfitz: I think you need to add `className` back to the const above, so it is not included in `props`, otherwise it overrides the `className` you are setting.

barelyfitz: And on second thought, it seems a little strange to spread `{...props}` onto an interior element rather than onto the top-most element of the component. I supposed this is being done because you need the `onClick` to be attached here to take advantage of the positioning?

barelyfitz: As an alternative, see #275. I think it doesn't add much extra code and moves the calculation of which star was clicked into this component. But I'm okay with either approach.

stories/theme/core/RatingBar.css
```diff
@@ -1,18 +1,84 @@
...
+
+.themedRating_star10 .RatingBar_Icons {
+  background-image: GoldStar;
+}
+
+.themedRating_star10 .RatingBar_Background {
+  background-image: GrayStar;
+  width: 200px;
+  height: 20px;
 }
```
barelyfitz: For the clickable Rating star, add `cursor: pointer;` style.

packages/react-ui-core/src/Ratings/RatingBar.js
tadjohnston: @barelyfitz you're correct about the `className`. It really should be added to the root node.  Also, by spreading `{...props}` it would override everything before it including `className`. So yeah, things you don't want to overwrite should be _after_ `{...props}`.

PR #270, opened 2018-05-03
fix(mapbox): reverting mapbox due to crashing

tadjohnston:

Mapbox is crashing after bumping to `0.44.2`. reverting to enforce an exact version of `0.42.2`


tadjohnston:
Don't have time, need this merged ASAP, hence merging.

PR #272, opened 2018-05-03
fix(mapbox): making a peer dependency

tadjohnston:

Changing mapbox-gl to a peer dependency so we have more control over it in the repo it's being used
in.


tadjohnston:
Closing this as we're just going to remove mapbox completely from react-ui-core.

PR #275, opened 2018-05-07
Ratings with optional onClick handler

barelyfitz:
Another option to consider for @ninomichaels Ratings component - adding click support directly into the component, but it only handles clicks if user passes in an `onClick` prop, otherwise avoids adding a click handler.


packages/react-ui-core/src/Ratings/RatingBar.js
```diff
@@ -58,6 +64,20 @@ export default class RatingBar extends PureComponent {
     )
   }
 
+  handleClick(event) {
```
tadjohnston: What does this need to be added for? Is this so you can click and select a rating? If not, I don't see a use case for this.

barelyfitz:
Closing since parent branch has been merged.

PR #276, opened 2018-05-07
fix(google maps): single marker needs map check

tadjohnston:

Marker component needs to check for the map existence. There was also a bug where I had changed the
name from `this.marker` to `this.markerInstance` and missed a couple places


packages/react-ui-ag/src/Gmap/Marker.js
```diff
@@ -1,7 +1,7 @@
 import React, { PureComponent } from 'react'
 import PropTypes from 'prop-types'
 import { Marker as Gmarker } from '@rentpath/react-ui-core'
-import { markerIcon } from './markerIcons'
+import { markerRedDotIcon } from './markerIcons'
```
bethstevic: Do the marker icons live in react-ui-ag now?

mpelzsherman: Can you help me understand the decision to change the icon names to "RedDot" and "BlackDot"? If we decide to change the marker color some day, won't this make it harder?

mpelzsherman:
LGTM, with small reservations about the icon names.

packages/react-ui-ag/src/Gmap/Marker.js
tadjohnston: @bethstevic there encoded in the urls now so we don't have to worry about them
@mpelzsherman I changed them because otherwise you can't look at them and easily tell what they are.  The logic here was that we may want to test other markers at some point and we can't do that unless we clearly specify the marker intent.  We could always add something to this like `markerBlueDotIcon`. Either way I'm fine with whatever we decide.

bethstevic:
Looks good :+1:

PR #281, opened 2018-05-07
fix(google map): lower anchor of selected icon

tadjohnston:

- Lowering position of selected icon as it's not covering the red dot marker completely
- Added
selected icon story


packages/react-ui-ag/src/Gmap/markerIcons.js
```diff
@@ -10,5 +10,5 @@ export const markerBlackDotIcon = () => ({
...
-  anchor: new window.google.maps.Point(19, 56),
+  anchor: new window.google.maps.Point(19, 54),
```
mpelzsherman: What do you think about making it possible to override the anchor coordinates in case this needs to change in the future?

tadjohnston: You can already do this by overriding the whole function if you need to. I was thinking about pulling these out of react-ui completely and just have the app handle it instead.

bethstevic:
Looks good to me 👍

PR #283, opened 2018-05-08
feat(react-ui-ag): add structured data for all listings

ostlerp:
affects: @rentpath/react-ui/core, @rentpath/react-ui-ag

- Create components for use in creating structured data schema
- Apply the structured data to available listings

### DIFF
--- a/packages/react-ui-core/src/Schema/Schema.js
+++ b/packages/react-ui-core/src/Schema/Schema.js
@@ -1,0 +1,110 @@
+import React, { PureComponent } from 'react'
+import PropTypes from 'prop-types'
+import themed from 'react-themed'
+
+const ApartmentComplex = ({ children, ...rest }) => (
+  <div itemType="https://schema.org/ApartmentComplex" itemScope {...rest}>
+    {children}
+  </div>
+)
+
+ApartmentComplex.propTypes = {
+  children: PropTypes.node,
+}
+
+const AggregateRating = ({ score, count, reviews }) => {
+  if (!count && !reviews) {
+    return null
+  }
+
+  return (
+    <div
+      itemScope
+      itemProp="aggregateRating"
+      itemType="https://schema.org/AggregateRating"
+    >
+      <meta itemProp="worstRating" content="1.0" />
+      <meta itemProp="bestRating" content="5.0" />
+      <meta itemProp="ratingValue" content={score} />
+      {!!count && <meta itemProp="ratingCount" content={count} />}
+      {!!reviews && <meta itemProp="reviewCount" content={reviews} />}
+    </div>
+  )
+}
+
+AggregateRating.propTypes = {
+  score: PropTypes.number,
+  count: PropTypes.number,
+  reviews: PropTypes.number,
+}
+
+const Address = ({ addressLine1, city, state, zip }) => (
+  <span itemType="https://schema.org/PostalAddress" itemProp="address" itemScope>
+    <meta itemProp="streetAddress" content={addressLine1} />
+    <meta itemProp="addressLocality" content={city} />
+    <meta itemProp="addressRegion" content={state} />
+    <meta itemProp="postalCode" content={zip} />
+  </span>
+)
+
+Address.propTypes = {
+  addressLine1: PropTypes.string,
+  city: PropTypes.string,
+  state: PropTypes.string,
+  zip: PropTypes.string,
+}
+
+const Geo = ({ latitude, longitude }) => {
+  if (!latitude || !longitude) {
+    return null
+  }
+
+  return (
+    <div itemProp="geo" itemScope itemType="https://schema.org/GeoCoordinates">
+      <meta itemProp="latitude" content={latitude} />
+      <meta itemProp="longitude" content={longitude} />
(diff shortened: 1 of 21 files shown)


packages/react-ui-core/src/Listing/ListingSchemas.js
```diff
@@ -0,0 +1,110 @@
...
+    return (
+      <div itemProp="name">
+        {url ?
+          <a
+            href={url}
+            itemProp="url"
+            target="_blank"
+            className={theme.PropertyLink}
+            onClick={event => event.preventDefault()}
+          >
```
mpelzsherman: Does this cause problems if you try to nest another `a` tag as a child?

packages/react-ui-core/src/Listing/ListingSchemas.js
```diff
@@ -0,0 +1,110 @@
...
+  children: PropTypes.node,
+}
+
+export default {
+  AddressSchema,
+  AggregateRatingSchema,
+  ApartmentComplexSchema,
+  GeoSchema,
+  NameAndUrlSchema,
+}
```
mpelzsherman: This file is getting a little big, but it's OK for now. If it gets much bigger I'd prefer to see it broken up, personally.

packages/react-ui-core/src/Listing/ListingSchemas.js
```diff
@@ -0,0 +1,110 @@
...
+  theme: PropTypes.object,
+  children: PropTypes.node,
+}
+
+export default {
+  AddressSchema,
+  AggregateRatingSchema,
+  ApartmentComplexSchema,
+  GeoSchema,
+  NameAndUrlSchema,
```
mpelzsherman: You could remove "Schema" from all of the schemas, since they're all "namespaced" under `ListingSchemas`.

packages/react-ui-core/src/Listing/ListingSchemas.js
ostlerp: Looks like it does! It probably won't come all that often though since the entire card is wrapped in an onClick so passing a link would be redundant anyway. The reason an anchor is added here is its better to tag structured data that's being rendered.

packages/react-ui-core/src/Listing/ListingSchemas.js
ostlerp: Yeah this is a tough one, I think. If they were imported directly as like an "ApartmentComplex" it might be pretty unclear what that component is trying to do.

packages/react-ui-core/src/Listing/ListingSchemas.js
tadjohnston: Is there any way to test this?

packages/react-ui-core/src/Listing/ListingSchemas.js
tadjohnston: I think if we create a root component dir that was `Schema` and exported them all from there, maybe then you could remove the suffix? I kind of agree the name is a little unnecessary and if you know that `import { ApartmentComplex } from '../Schema'` I think it's safe to assume where it came from 😄 .

packages/react-ui-core/src/Schema/Schema.js
```diff
@@ -0,0 +1,110 @@
...
+}
+
+@themed(['PropertyLink'], { pure: true })
+class NameAndUrl extends PureComponent {
+  render() {
+    const { url, theme, children } = this.props
+    return (
+      <div itemProp="name">
+        {url ?
+          <a
```
tadjohnston: Does this have to be anchor tag? Or can we append it to the div tag instead? Or vice versa?

ostlerp:
Updated naming conventions to remove the "Smurf" suffixes.

packages/react-ui-core/src/Schema/Schema.js
ostlerp: Confirmed; needs to be an anchor =\

tadjohnston: Can the `itemProp="name"` be merged into the anchor tag?

tadjohnston: We want to have the least amount of dom nodes as possible as it pertains to a listing.

ostlerp: Ah, I understand the concern now. So the problem is the conflicting `itemProp`s that need to be on separate tags. I'm definitely open to suggestions as I'm not sure how to get around having the two nodes here.

tadjohnston: Can you just do:
```
if (url) {
  return <a itemProp="name" href.... >{c
}

return <div itemProp="name">{children}</div>
```

tadjohnston: I guess my question would if it has a url does it also need an `itemProp="name"`

ostlerp: It needs both in this case. This is the ideal scenario: 
screen shot 2018-05-14 at 3 27 16 pm

PR #284, opened 2018-05-08
fix(google maps): check for valid projection from Google Maps API

tcwilson:


tcwilson:
More info: https://stackoverflow.com/questions/17191664/why-getprojection-is-not-working-in-v3

Really hard to reproduce this issue. I think the props are somehow changing on rent-js and `componentDidUpdate` is called before the map is fully initialized and calls `draw()` itself.

PR #285, opened 2018-05-09
fix(google maps): loader race condition

tadjohnston:

Fixing google script loading race condition where 2 google map api scripts can't be on the page at
the same time

To replicate currently you have to add a `@rentpath/react-ui-ag` `PdpMap` to the PDP card
1. click on a listing on mobile srp
2. pdp card loads on srp route
3. gmaps gets mounted 
4. route changes to pdp (without reload)
5. pdp card loads on pdp route
6. gmaps gets mounted again

The amount of time in between routes causes a race condition where the gmaps api script loader thinks it doesn't have a script loaded yet so it tries to load another. The first one that tried to load on SRP is now unmounted so a global callback no longer works.  The only way seemingly to get this working properly is to add a global listener (and yes, I tried adding a global var, but you have the same issue).

### DIFF
--- a/packages/react-ui-core/src/Gmap/withGoogleScript.js
+++ b/packages/react-ui-core/src/Gmap/withGoogleScript.js
@@ -5,6 +5,8 @@ import { parseArgs } from '@rentpath/react-ui-utils'
 import getDisplayName from './utils/getDisplayName'
 
 const API_BASE_URL = 'https://maps.googleapis.com/maps/api/js'
+const SCRIPT_ID = 'google-maps-api-script'
+const EVENT_NAME = 'googleMapsLoaded'
 
 export default function(BaseComponent) {
   class Container extends PureComponent {
@@ -22,6 +24,7 @@ export default function(BaseComponent) {
         PropTypes.object,
       ]),
       onScriptLoadError: PropTypes.func,
+      loaded: PropTypes.bool,
     }
 
     static defaultProps = {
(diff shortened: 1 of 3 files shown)


tcwilson:
👍 
I'll give this a try on rent-js.

bethstevic:
I don't fully understand all that it is doing, but as long as it works the code looks good, doesn't look like it would introduce any bugs, so 👍 from me

packages/react-ui-core/src/Gmap/withGoogleScript.js
```diff
@@ -33,20 +36,31 @@ export default function(BaseComponent) {
...
       spinner: this.props.spinner,
     }
 
     componentDidMount() {
-      if (this.isInitialized) {
-        this.scriptLoaded()
+      if (this.state.loaded || (this.hasExistingScript && this.isInitialized)) {
+        // no other place to put this due to SSR
+        this.setState({ loaded: true }) // eslint-disable-line react/no-did-mount-set-state
         return
```
lfender6445: is the return necessary?

lfender6445:
👍

packages/react-ui-core/src/Gmap/withGoogleScript.js
tadjohnston: Yes, because we don't need any of the following to be called if the api has fully loaded.

PR #289, opened 2018-05-10
fix(google maps): switch to CustomEvent, so polyfill on IE11 will work

tcwilson:


tadjohnston:
One line fix, merge it please 👍

PR #292, opened 2018-05-11
fix(@rentpath/react-ui-core): Gmaps Prototype checking

orionstein:
Has doesn't work with gmaps prototype functions, and needed to check with get


lfender6445:
👍

PR #293, opened 2018-05-11
feat(GMAP): overlay Positioning

dawsonpa:

Add Prop to the GMAP OverlayView called overlayPosition. It lets you set the absolute position of
the overlay on the map. If it is not provided then the default positioning is used.

### DIFF
--- a/packages/react-ui-core/src/Gmap/OverlayView.js
+++ b/packages/react-ui-core/src/Gmap/OverlayView.js
@@ -121,7 +121,7 @@ export default class OverlayView extends PureComponent {
       // Projection is undefined if map has not yet initialized
       if (!projection) return
 
-      const { x, y } = this.overlay.getProjection().fromLatLngToDivPixel(position)
+      const { x, y } = projection.fromLatLngToDivPixel(position)
 
       // Show overlay only when it is in view.
       if (Math.abs(x) < 4000 && Math.abs(y) < 4000) {
--- a/packages/react-ui-core/src/Gmap/OverlayView.js
+++ b/packages/react-ui-core/src/Gmap/OverlayView.js
@@ -13,6 +13,10 @@ export default class OverlayView extends PureComponent {
     preventBubbleEvents: PropTypes.bool,
     className: PropTypes.string,
     theme: PropTypes.object,
+    overlayPosition: PropTypes.shape({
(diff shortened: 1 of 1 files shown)


packages/react-ui-core/src/Gmap/OverlayView.js
```diff
@@ -111,18 +116,19 @@ export default class OverlayView extends PureComponent {
     let display = 'none'
 
     if (position) {
-      const projection = this.overlay.getProjection()
-
-      // Projection is undefined if map has not yet initialized
-      if (!projection) return
```
mpelzsherman: Why was this removed?

packages/react-ui-core/src/Gmap/OverlayView.js
```diff
@@ -111,18 +116,19 @@ export default class OverlayView extends PureComponent {
...
-
-      const { x, y } = projection.fromLatLngToDivPixel(position)
+      const { x, y } = this.overlay.getProjection().fromLatLngToDivPixel(position)
```
mpelzsherman: Without the above check, this like could throw if `this.overlay.getProjection()` is undefined or null.

packages/react-ui-core/src/Gmap/OverlayView.js
dawsonpa: Yea good catch. It should not have been removed.

packages/react-ui-core/src/Gmap/OverlayView.js
dawsonpa: True

packages/react-ui-core/src/Gmap/OverlayView.js
```diff
@@ -116,13 +121,19 @@ export default class OverlayView extends PureComponent {
       // Projection is undefined if map has not yet initialized
       if (!projection) return
 
-      const { x, y } = projection.fromLatLngToDivPixel(position)
+      const { x, y } = this.overlay.getProjection().fromLatLngToDivPixel(position)
```
mpelzsherman: why not just use `projection` here? no need to call `this.overlay.getProjection()` again... is there?

tadjohnston:
Closing, not needed anymore.

PR #295, opened 2018-05-11
chore(versioning): fixing broken versioning

tadjohnston:
affects: @rentpath/react-ui-ag, @rentpath/react-ui-rent, @rentpath/react-ui-rentals

Need to major bump to address already published package issues


orionstein:
👍

PR #297, opened 2018-05-12
fix(Broaden event blacklist for overlay): fixes issue where ipad taps…

lfender6445:
Dont close the annotation window for IOS taps


tcwilson:
👍

PR #300, opened 2018-05-15
ModalStack port to react-ui

orionstein:
Moving ModalStack back to react-ui, so both Rent and AG can use this

Added a comment, but will note here, because the input format might want to be talked about -

Expects currentModal to come in as an object that looks like
```
{
  id: modalId,
  props: { ...modalProps },
}
```
Expects modalDefinitions to come in as an object that looks like
```
{
  [modalId]: {
    name: modalName,
    overlay: isOverlayOpen,
    resolve: () => import(modalPath),
  }
}
```

### DIFF
--- a/packages/react-ui-core/src/ModalStack/__tests__/__snapshots__/ModalStack-test.js.snap
+++ b/packages/react-ui-core/src/ModalStack/__tests__/__snapshots__/ModalStack-test.js.snap
@@ -1,0 +1,5 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`ModalStack renders a modal correctly 1`] = `null`;
+
+exports[`ModalStack renders correctly 1`] = `null`;
--- a/packages/react-ui-ag/src/Listings/__tests__/DesktopListing-test.js
+++ b/packages/react-ui-ag/src/Listings/__tests__/DesktopListing-test.js
@@ -20,7 +20,6 @@ const baseListing = {
   rating: {
     score: 4,
     label: '20',
-    uniqueId: '111',
     count: 41,
     reviews: 2,
   },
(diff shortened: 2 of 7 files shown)


orionstein:
We can hook it up in rent like 

```
import { connect } from 'react-redux'
import { currentModal as currentModalSelector, makeGetModalDefinition } from 'app/selectors/modal'
import { closeModal } from 'app/store/shared/actions'
import { ModalStack } from '@rentpath/react-ui'

const makeMapStateToProps = (_, initialProps) => {
  const largeGetModalDefinition = makeGetModalDefinition(initialProps.modalDefinitions)
  return state => ({
    currentModal: currentModalSelector(state),
    modalDefinition: largeGetModalDefinition(state),
  })
}

const mapDispatchToProps = {
  closeModal,
}

export default connect(makeMapStateToProps, mapDispatchToProps)(ModalStack)
```

orionstein:
Updated it to just take in all the modalDefinitions, and use those to drive the modal. 

This way, we pass in the modalDefinitions for the app section, and this can handle loading them, displaying them, and closing them

lfender6445:
looks good but needs some tests

mbrimmer83:
LG other than tests 👍

packages/react-ui-core/src/ModalStack/__tests__/__snapshots__/ModalStack-test.js.snap
```diff
@@ -0,0 +1,5 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`ModalStack renders a modal correctly 1`] = `null`;
+
+exports[`ModalStack renders correctly 1`] = `null`;
```
lfender6445: ehhh

orionstein: Yeh

PR #304, opened 2018-05-17
Ml/grey inactive pin

WinstonMain:
[story](https://rentpath.leankit.com/card/654972651)

Support grey pins for inactive properties for AG


packages/react-ui-ag/src/Gmap/Markers.js
```diff
@@ -19,30 +20,33 @@ export default class Markers extends PureComponent {
...
-        if (onMouseOver) onMouseOver(marker)
-      },
-      onMouseOut: (event, props, marker) => {
-        marker.setIcon(markerIcon())
-        if (onMouseOut) onMouseOut(marker)
-      },
-      ...this.props.marker(feature),
-    })
+    return feature => {
+      const defaultMarker = (feature.properties.isActive) ? redDotIcon : greyDotIcon
```
tadjohnston: Should read:
```
const defaultMarker = !feature.properties.isActive ? greyDotIcon : redDotIcon
```

in the event that `isActive` isn't there.

lfender6445:
👍

tadjohnston:
No wip anymore @lfender6445 removing it.

packages/react-ui-ag/src/Gmap/__tests__/Markers-test.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react'
 import { shallow } from 'enzyme'
 import Markers from '../Markers'
-import { markerRedDotIcon, markerBlackDotIcon } from '../markerIcons'
+import { redDotIcon, blackDotIcon, blackDotIconWithBalloon } from '../markerIcons'
```
tadjohnston: A test needs to be aded for the inactive listing icon.

packages/react-ui-ag/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@rentpath/react-ui-ag",
-  "version": "5.1.1",
+  "version": "6.0.0",
```
tadjohnston: Please don't touch this file.

packages/react-ui-ag/src/Gmap/Marker.js
```diff
@@ -16,7 +15,6 @@ export default class Marker extends PureComponent {
 
   get marker() {
     return props => ({
-      icon: markerRedDotIcon(),
```
tadjohnston: Why did you remove this?

packages/react-ui-ag/src/Gmap/__tests__/Markers-test.js
```diff
@@ -1,30 +1,49 @@
...
       const wrapper = shallow(<Markers map={map} />)
       expect(wrapper.instance().marker()).toEqual({
-        icon: markerRedDotIcon(),
+        icon: redDotIcon(),
         onMouseOver: expect.any(Function),
         onMouseOut: expect.any(Function),
       })
     })
+    it('uses a greyDotIcon for inactive properties', () => {
+      const markers = new Markers()
```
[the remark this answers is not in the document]
WinstonMain: How can you pass `feature` if you shallow render `Markers`? It should be sufficient to invoke the getter from the class instance and assert that the logic is correct.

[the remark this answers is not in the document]
tadjohnston: I mean I guess you could to is this way. The normal way would be something like this:
```
const wrapper = shallow(<Markers map={map} markerIconHover={NOOP} />)
const instance = wrapper.instance()
const icon = greyDotIcon()

expect(instance.marker(feature)). toContain({ icon })
```
(haven't tested this, but something like this would work)

This will actually matter if you make a default `markertIconInactive` prop.

packages/react-ui-ag/src/Gmap/Marker.js
mpelzsherman: He just renamed it. I like the new name. Less Smurfy.

mpelzsherman:
Retroactive 👍

packages/react-ui-ag/src/Gmap/Marker.js
tadjohnston: It was removed before from what I saw, I dunno 🤷‍♂️

PR #308, opened 2018-05-21
feat(react-ui-core Markers): add an index to Markers.

mpelzsherman:

In order to scroll the selected listing into view, an index is required in order to compute the page
that needs to be loaded.

### DIFF
--- a/packages/react-ui-core/src/Gmap/Markers.js
+++ b/packages/react-ui-core/src/Gmap/Markers.js
@@ -90,12 +90,12 @@ export default class Markers extends PureComponent {
 
     if (!map || !geojson || !geojson.features) return
 
-    geojson.features.forEach(feature => {
+    geojson.features.forEach((feature, index) => {
       const id = feature.properties.id
 
       if (this.markers[id]) return
 
-      this.markers[id] = setupMarker(map, this.marker(feature))
+      this.markers[id] = setupMarker(map, this.marker(feature), index)
     })
   }
 
--- a/packages/react-ui-core/src/Gmap/utils/markerHelpers.js
+++ b/packages/react-ui-core/src/Gmap/utils/markerHelpers.js
@@ -32,12 +32,13 @@ export const removeMarker = marker => {
   }
 }
 
-export const setupMarker = (map, props) => {
+export const setupMarker = (map, props, index) => {
   if (map) {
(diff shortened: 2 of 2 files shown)


tadjohnston:
I'm not seeing any sort of index usage in the ag.js PR.  Can you explain what the correlation is between the 2 PRs?

packages/react-ui-core/src/Gmap/Markers.js
```diff
@@ -90,12 +90,12 @@ export default class Markers extends PureComponent {
...
     if (!map || !geojson || !geojson.features) return
 
-    geojson.features.forEach(feature => {
+    geojson.features.forEach((feature, index) => {
       const id = feature.properties.id
 
       if (this.markers[id]) return
 
-      this.markers[id] = setupMarker(map, this.marker(feature))
+      this.markers[id] = setupMarker(map, this.marker(feature), index)
```
barelyfitz: So you have a set of items in `geojson`, you loop through them and save a marker based on the id, and you save the geojson `index` of the item.

What happens if `geojson` changes, but the new array contains some of the same items? It seems like the previous marker would still exist and have an `index` value that is no longer correct?

barelyfitz: Also does that mean that this is no longer a pure component?

nray621: + 1 for Fitz's first point here

mpelzsherman: When the `geojson` changes, we render a completely new set of Markers.

I don't understand why this change would make this component "impure".

mpelzsherman:
After discussing with @orionstein I'm going to write a selector to derive the index instead.

packages/react-ui-core/src/Gmap/Markers.js
barelyfitz: For example:
* I call the component with geojson ids [ a, b, c ]
* It loops through and create this.markers = { a: index0, b: index1, c: index2 }
* I update the component with geojson ids [ b, d ]
* It loops through and sees that this.markers.b already exists so it does not create a new marker, and creates {d: index1}
* So would I end up with this.markers = { b: index1, d: index1 }

PR #309, opened 2018-05-21
fix(react-image-gallery): true lazy loading

tadjohnston:

Forked the repo to address: https://github.com/xiaolin/react-image-gallery/issues/278

Gallery: https://github.com/xiaolin/react-image-gallery/compare/master...rentpath:master


tadjohnston:
I think the only thing I'd worry about is the thumbnails, but I don't think we use them anywhere.  However, they should still work.

PR #310, opened 2018-05-21
fix(react-ui-core): modalStack default export

orionstein:

Fixes a bug when we pass a Modal to Modalstack as the default export.
Cleans this up a little bit,
and adds some more error handling


lfender6445:
👍

PR #313, opened 2018-05-22
fix(react-ui-ag): revert markerIcons to fix colors being changed

nray621:

Fill colors were changed for markerIcons - changing back
Change version to v in google maps script component


tadjohnston:
Merging this as it's critical.

PR #315, opened 2018-05-22
feat(react-ui-ag): Modify MobileListing and its stories, tests, theme

nray621:


packages/react-ui-ag/src/Listings/MobileSRPListing.js
```diff
@@ -0,0 +1,82 @@
...
+import Listing from './Listing'
+
+@themed([
+  'MobileSRPListing',
+  'BedsAndBaths',
+  'UnitLevelAvailabilityAndLastUpdated',
+  'LastUpdated',
+  'Phone',
+], { pure: true })
+export default class MobileSRPListing extends PureComponent {
```
tadjohnston: Can we call this something that doesn't have SRP in the name? If it's something that can be reused across mobile how about just `MobileListing`?

packages/react-ui-ag/src/Listings/MobileSRPListing.js
```diff
@@ -0,0 +1,82 @@
+import React, { PureComponent } from 'react'
```
tadjohnston: Need tests for this.

packages/react-ui-ag/src/Listings/MobileSRPListing.js
nray621: Yeah no problem. There is already a generic `MobileListing`, but I'm confused because I don't think we actually use it anywhere

packages/react-ui-ag/src/Listings/MobileSRPListing.js
nray621: Yup. I just went ahead and got a PR up, but this is very much a work in progress

nray621:
@ostlerp Oops missed that. Yeah we can definitely remove that. I'll do it now

PR #318, opened 2018-06-04
Deleted Grid components from React UI Core and Storybook

davidthrasher:


lfender6445:
👍 after this goes through, please update your rent-js branch so that it has the new version of react-ui-core

PR #320, opened 2018-06-06
Add onPhotoCarouselSlide prop

justinrex:
[Card](https://rentpath.leankit.com/card/674996770)

Add ability to pass in a function to be called when carousel is slid

### DIFF
--- a/packages/react-ui-ag/src/Listings/Listing.js
+++ b/packages/react-ui-ag/src/Listings/Listing.js
@@ -4,6 +4,7 @@ import themed from 'react-themed'
 import classnames from 'classnames'
 import autobind from 'autobind-decorator'
 import get from 'lodash/get'
+import isEqual from 'lodash/isEqual'
 import LazyLoad, { forceCheck } from 'react-lazyload'
 import { Button, ToggleButton, ListingComponents, ListingCell } from '@rentpath/react-ui-core'
 import { Banner } from '../Banners'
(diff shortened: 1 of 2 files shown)


packages/react-ui-ag/src/Listings/Listing.js
```diff
@@ -135,13 +140,7 @@ export default class Listing extends Component {
...
   renderCtaButton(props, key) {
     const { theme, listing } = this.props
-    const {
-      className,
-      onClick,
-      valueLocation,
-      children,
-      ...rest
-    } = props
+    const { className, onClick, valueLocation, children, ...rest } = props
```
tadjohnston: I personally thought these were easier to read on newlines, but to each their own i guess.

justinrex: I just have prettier format things on save. I can undo if you'd like. I don't have a preference.

LinnJS: I think if you would of added one more item it would have broken it out onto their own lines lol

justinrex: I've seen prettier in some strange case put all but one prop on the same line, that one I really disagree with.

tadjohnston: Well, we have an eslint requirement in ag.js if it's 4 or more props, put them on new lines.  We will most likely be moving to prettier anyway, so I'm sure we can discuss it then.  @LinnJS I wonder if this has to do with max line length for breaking them into new lines?

PR #322, opened 2018-06-12
feat(gmap): layer grey markers under red markers

tamars2:

Conditionally apply zIndex to markers based on `isActive`

### DIFF
--- a/packages/react-ui-ag/src/Gmap/Markers.js
+++ b/packages/react-ui-ag/src/Gmap/Markers.js
@@ -51,6 +51,7 @@ export default class Markers extends PureComponent {
           if (onMouseOut) onMouseOut(marker)
         },
         ...this.props.marker(feature),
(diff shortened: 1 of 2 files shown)


packages/react-ui-ag/src/Gmap/Markers.js
```diff
@@ -51,6 +51,7 @@ export default class Markers extends PureComponent {
           if (onMouseOut) onMouseOut(marker)
         },
         ...this.props.marker(feature),
+        zIndex: isActive ? 2 : 1,
```
tadjohnston: Can you add a test for this? Basically check to make sure zIndex is 2 when active, 1 when inactive.

tamars2: You're right.  I edited and added tests to be more explicit.

PR #323, opened 2018-06-12
refactor MobileListing component to allow for modifications

bethstevic:
[Card](https://rentpath.leankit.com/card/679902158)

### DIFF
--- a/packages/react-ui-ag/src/Listings/MobileListing.js
+++ b/packages/react-ui-ag/src/Listings/MobileListing.js
@@ -37,18 +37,22 @@ export default class MobileListing extends PureComponent {
       <div className={theme.MobileListing_LastUpdated}>{lastUpdated}</div>
   }
 
-  get renderInfo() {
+  get renderTopComponents() {
     const { theme, listing, ratings, sponsoredText } = this.props
     const { singleFamily, rating, sponsored } = listing
+    return !singleFamily && (
+      <div className={theme.MobileListing_RatingAndSponsored}>
+        {rating && <ListingComponents.Ratings data-tid="ratings" {...ratings} />}
+        {sponsored && <Text className={theme.MobileListing_Sponsored}>{sponsoredText}</Text>}
+      </div>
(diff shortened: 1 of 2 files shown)


packages/react-ui-ag/src/Listings/MobileListing.js
```diff
@@ -38,9 +39,37 @@ export default class MobileListing extends PureComponent {
...
+    if (removeCtaButtons) {
+      return (
+        <React.Fragment>
+          {!singleFamily &&
+            <div className={theme.MobileListing_RatingAndSponsored}>
+              {rating && <ListingComponents.Ratings data-tid="ratings" {...ratings} />}
+              {sponsored && <Text className={theme.MobileListing_Sponsored}>{sponsoredText}</Text>}
+            </div>
+          }
+          <div className={theme.MobileListing_Info_Left_Side}>
```
tadjohnston: Can't all of this be done through css instead of doing this here?

packages/react-ui-ag/src/Listings/MobileListing.js
```diff
@@ -13,6 +13,7 @@ export default class MobileListing extends PureComponent {
     className: PropTypes.string,
     ratings: PropTypes.object,
     sponsoredText: PropTypes.string,
+    removeCtaButtons: PropTypes.bool,
```
tadjohnston: Is this really needed? Doesn't the `ctaButtons` check in Listing do enough?

packages/react-ui-ag/src/Listings/MobileListing.js
```diff
@@ -49,6 +48,15 @@ export default class MobileListing extends PureComponent {
             {sponsored && <Text className={theme.MobileListing_Sponsored}>{sponsoredText}</Text>}
           </div>
         }
+      </React.Fragment>
```
orionstein: Do we need a fragment on this one since we return a single div wrapper for now?

PR #330, opened 2018-06-15
feat(photos): adding photo component

tadjohnston:

Adding a new photo component to provide beter support
* add BackgroundPhoto component
* add
Photo component
* switch carousel and other components to use these


packages/react-ui-core/src/Photo/BackgroundPhoto.js
```diff
@@ -0,0 +1,35 @@
+import React, { PureComponent } from 'react'
+import PropTypes from 'prop-types'
+import classnames from 'classnames'
+import themed from 'react-themed'
+
+@themed(['BackgroundPhoto'], {
+  pure: true,
+})
+
+export default class Photo extends PureComponent {
```
orionstein: This one should be BackgroundPhoto

packages/react-ui-core/src/Photo/BackgroundPhoto.js
```diff
@@ -0,0 +1,35 @@
...
+  static defaultProps = {
+    theme: {},
+  }
+
+  render() {
+    const { theme, className, url, alt, ...rest } = this.props
+
+    return (
+      <div
+        style={{ backgroundImage: `url(${url})` }}
```
orionstein: If we want to allow for a generic fallback, we can use something like

`
    style={{ backgroundImage: `url(${url})${fallbackUrl ? `,url(${fallbackUrl})` : ''}` }}
`

packages/react-ui-core/src/Photo/Photo.js
```diff
@@ -0,0 +1,35 @@
...
+  static defaultProps = {
+    theme: {},
+  }
+
+  render() {
+    const { theme, className, url, ...rest } = this.props
+
+    return (
+      <img
+        src={url}
```
orionstein: If we use a fallbackUrl like above, we could use onError here to accomplish it in the image

packages/react-ui-core/src/Photo/Photo.js
```diff
@@ -0,0 +1,62 @@
...
+
+  fallback() {
+    this.setState({ error: true })
+  }
+
+  render() {
+    const {
+      theme,
+      className,
+      url: _,
```
lfender6445: interesting, is this so that its omitted from rest?

packages/react-ui-core/src/Carousel/PhotoCarousel.js
```diff
@@ -29,12 +30,15 @@ export default class PhotoCarousel extends PureComponent {
...
     theme: {},
   }
 
-  photo(path) {
+  photo(item) {
     const { server, dimensions } = this.props
-    return `${server}${path}${dimensions}`
+    const { url, path, id } = item
+
+    return url || `${server}${path || id}${dimensions}`
```
lfender6445: why support path and id?

packages/react-ui-core/src/Photo/Photo.js
tadjohnston: Yep 😄 , just renaming it to `_`

packages/react-ui-core/src/Carousel/PhotoCarousel.js
tadjohnston: Well, AG is currently using `id`.  I'm not opposed to removing it, but I'm assuming your path is something like `imgr/12345` or no?  If not then I guess they're synonymous?  If it's using `imgr` we'd either have to convert the bff or map over every photo to change the key name.

lfender6445:
👍

packages/react-ui-core/src/Carousel/CarouselNavigation.js
```diff
@@ -21,6 +21,11 @@ export default class CarouselNavigation extends PureComponent {
     theme: {},
   }
 
+  get tagItem() {
+    const direction = this.props.direction === 'previous' ? 'left' : 'right'
+    return `${direction}_arrow`
```
mpelzsherman: Can this be overridden if needed?

tadjohnston: Yes... but why would you need to?

tadjohnston: screen shot 2018-06-18 at 4 03 18 pm

You should probably look at the code a little more on how things are done.  I can help explain though.  The structure for all react-ui components is generally setup up to spread the extra props over the component. That keeps for flexible / reusable code.  If you look at the above the spread props are _after_ the `data-tag_item` which means that if you were to pass `data-tag_item="foo"` in as a prop, the component would now have that on the `<button data-tag_item="foo">` DOM node since it gets passed through to `<Button />` and Button [spreads props](https://github.com/rentpath/react-ui/blob/master/packages/react-ui-core/src/Button/Button.js#L37) as well.

packages/react-ui-core/src/Photo/Photo.js
```diff
@@ -0,0 +1,62 @@
...
+      fallbackUrl,
+      ...rest
+    } = this.props
+
+    const { url, error } = this.state
+
+    /* eslint-disable jsx-a11y/alt-text */
```
barelyfitz: I assume you can pass `alt` as one of the "rest" params. Since that is a lint rule in the first place, what is our policy of always specifying an `alt` attribute? Should it be specified a prop and encourage to be used?

packages/react-ui-core/src/Photo/BackgroundPhoto.js
```diff
@@ -0,0 +1,43 @@
...
+import themed from 'react-themed'
+
+@themed(['BackgroundPhoto'], {
+  pure: true,
+})
+
+export default class BackgroundPhoto extends PureComponent {
+  static propTypes = {
+    url: PropTypes.string,
+    fallbackUrl: PropTypes.string,
```
barelyfitz: Looking at the definition of `background-image` style, it seems that when you specify multiple URLs it layers one on top of the other: [Using_multiple_backgrounds](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Backgrounds_and_Borders/Using_multiple_backgrounds)

So the term "fallback" seems a little misleading, and could lead to some unintended problems:
* If the image and the fallback image are different sizes, the fallback image might show up behind the main image.
* If the main image contains transparency, the fallback image might show through the transparent portions of the main image
* Both the main image and the fallback image will always be loaded and rendered to the page by the browser

Considering all the above it might be worthwhile adding some comments / warnings, to ensure this is used only in situations where it will not cause problems.

packages/react-ui-core/src/Photo/BackgroundPhoto.js
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  get fallback() {
+    const { fallbackUrl } = this.props
+    return fallbackUrl ? `url(${fallbackUrl})` : undefined
+  }
+
+  render() {
+    const { theme, className, url, alt, ...rest } = this.props
+    const image = [`url(${url})`].concat(this.fallback).filter(Boolean).join()
```
barelyfitz: This line took me a couple minutes to figure out what it was doing (because I'm a little slooow on the uptake).

Maybe split apart into something more readable?
```
joinUrlStyles(...urls) {
  return urls
    .filter(Boolean) // skip if undefined
    .map(url => `url(${url})`)
    .join() // defaults to comma-separated
```

Then `joinUrlStyles(url, this.fallback)`?

packages/react-ui-core/src/Photo/Photo.js
```diff
@@ -0,0 +1,62 @@
...
+    return (
+      <img
+        src={error ? fallbackUrl : url}
+        className={classnames(
+          className,
+          theme.Photo
+        )}
+        onError={!error && this.fallback}
```
barelyfitz: Nice, looks like this will not get caught in any type of a loop. And if you update the url, the new url can still have a fallback.

One question I have is around performance - since this is a stateful component, are there any concerns if we have, say, 100 of these components on the page?

packages/react-ui-core/src/Photo/Photo.js
tadjohnston: Yeah, I was thinking maybe we just want to enforce it. 🤷‍♂️

packages/react-ui-core/src/Photo/BackgroundPhoto.js
tadjohnston: Yeah, I hear you. The reason this is called `fallbackUrl` was to match with the Photo component. That way you don't have to do different things to essentially pass the same logic to each of them.

packages/react-ui-core/src/Photo/BackgroundPhoto.js
tadjohnston: I can do that if you'd like 😄

orionstein: More like

```
const joinUrlStyles = pipe(
  (...args) => ([...args]),
  filter(Boolean),
  map(url => `url(${url})`),
  join(','),
)
```

`joinUrlStyles(url, this.fallback)`

packages/react-ui-core/src/Photo/Photo.js
tadjohnston: How else do you propose we do it? I guess we could try to remove state, but it might be hacky 🤷‍♂️

tadjohnston:
I went a little nuts on package updates...should have probably done it in another PR.

packages/react-ui-core/src/Photo/Photo.js
```diff
@@ -0,0 +1,93 @@
...
+    fallbackUrl: PropTypes.string,
+    theme: PropTypes.object,
+    className: PropTypes.string,
+  }
+
+  static defaultProps = {
+    theme: {},
+  }
+
+  // don't need the contructdor once we fix the getDerivedStateFromProps
```
barelyfitz: Spelling: "contructdor"

grenti:
LGTM - should we add a story for the `Photo` and `BackgroundPhoto` components?

tadjohnston:
We _can_ add a story. We can probably add it after this PR as I'd like to get this in. This one has been a long time coming 😄

packages/react-ui-core/src/Photo/Photo.js
nray621: I share Fitz's concerns about it being stateful in general, but it doesn't look like we're in danger of triggering any useless re-renders as far as I can tell.

tadjohnston: How about something like this instead: https://gist.github.com/tadjohnston/b7fa785b4e90f551f6a634908e5c3d35

PR #331, opened 2018-06-15
fix(@rentpath/react-ui-ag): Pass renderCustomControls through Listing

justinrex:

fixes issue where Listing was not passing `renderCustomControls` through to Carousel


packages/react-ui-ag/src/Listings/Listing.js
```diff
@@ -189,7 +190,7 @@ export default class Listing extends Component {
   }
 
   renderPhotoCarousel() {
-    const { listing, theme, photos, isActive } = this.props
+    const { listing, theme, photos, isActive, renderCustomControls } = this.props
```
LinnJS: 👍

PR #332, opened 2018-06-18
Allow MobileMapListing to display text on no match

nray621:
feat(react-ui-ag): Allow MobileMapListing to display text on no match

### DIFF
--- a/packages/react-ui-ag/src/Listings/MobileListing.js
+++ b/packages/react-ui-ag/src/Listings/MobileListing.js
@@ -12,7 +12,6 @@ export default class MobileListing extends PureComponent {
     theme: PropTypes.object,
     className: PropTypes.string,
     ratings: PropTypes.object,
-    sponsoredText: PropTypes.string,
     categoryMatch: PropTypes.bool,
     noMatchText: PropTypes.string,
   }
@@ -21,7 +20,6 @@ export default class MobileListing extends PureComponent {
     theme: {},
     listing: {},
     ratings: {},
-    categoryMatch: true,
   }
 
   get availabilityOrUpdated() {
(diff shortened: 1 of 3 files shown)


packages/react-ui-core/src/Listing/ListingComponents.js
```diff
@@ -66,9 +66,13 @@ Ratings.contextTypes = {
   rating: PropTypes.object,
 }
 
-const Photos = (props, { photos }) => (
-  photos ? <PhotoCarousel {...props} items={photos} /> : null
-)
+const Photos = (props, { photos }) => {
```
nray621: This looks to do exactly the same thing. Honestly just curious why this is preferable to the ternary

stories/theme/ag/small/MobileListing.css
```diff
@@ -23,6 +23,22 @@
...
+  max-width: 65%;
+  margin: 0 auto;
+  background: rgba(0, 0, 0, 0.5);
+  color: #FFFFFF;
+  border-radius: 5px;
+  padding: 0 6px;
+  text-align: center;
+  font-size: 12px;
+  font-weight: 700;
+  left: 50%;
```
nray621: Not a big deal at all, but it'd be nice if we logically group attributes to make it easier to read. 
ex - position followed by top followed by left etc.

packages/react-ui-ag/src/Listings/MobileListing.js
```diff
@@ -41,14 +39,19 @@ export default class MobileListing extends PureComponent {
...
-        {!categoryMatch && <div className={theme.MobileListing_NoMatch}>{noMatchText}</div>}
-        {rating && <ListingComponents.Ratings data-tid="ratings" {...ratings} />}
-        {sponsored && <Text className={theme.MobileListing_Sponsored}>{sponsoredText}</Text>}
-      </div>
+
+    return (
+      <React.Fragment>
+        {noMatchText && <Text className={theme.MobileListing_NoMatchText}>{noMatchText}</Text>}
+        {!singleFamily &&
+          <div className={theme.MobileListing_RatingAndSponsored}>
```
nray621: I hate to add more conditionals here, but it looks like the `MobileListing_RatingAndSponsored` div is going to be rendered regardless of either of its children existing. I might be wrong, but from what I've seen its fairly common for neither to exist, and we could shave a off a DOM node.

packages/react-ui-ag/src/Listings/MobileListing.js
```diff
@@ -21,7 +20,6 @@ export default class MobileListing extends PureComponent {
     theme: {},
     listing: {},
     ratings: {},
-    categoryMatch: true,
```
nray621: 👍

nray621:
I approve of this code. LOL. I can't actually approve since it was originally my PR though, which is appropriate. So it'll need one more review

bethstevic:
Looks good to me

PR #334, opened 2018-06-19
Bs/668249505/free draw core component

yasabere:
* handles incomplete shapes - close shape with a straight line
* Returns point data on shape complete
* Renders placeholder control to test functionality


packages/react-ui-core/src/Gmap/FreeDrawLayer.js
```diff
@@ -0,0 +1,92 @@
+import React, { PureComponent } from 'react'
+import PropTypes from 'prop-types'
+import autobind from 'autobind-decorator'
+import themed from 'react-themed'
+
+@themed(['FreeDraw', 'FreeDraw_ToolTip'], { pure: true })
```
barelyfitz: I only see reference to 'FreeDraw_Tooltip'.
Since this style is in the `Gmap.css` perhaps it would be better as `Gmap_FreeDrawTootip`

packages/react-ui-core/src/Gmap/FreeDrawLayer.js
```diff
@@ -0,0 +1,92 @@
...
+
+  render() {
+    const { theme } = this.props
+
+    return (
+      <button
+        className={theme.FreeDraw_ToolTip}
+        onClick={this.enableMapDraw}
+      >
+        Click to draw
```
barelyfitz: Make this a prop with a default value?

stories/theme/core/Gmap.css
```diff
@@ -23,3 +23,8 @@
   padding: 5px;
   white-space: nowrap;
 }
+
+.FreeDraw_ToolTip {
+  position: absolute;
+  z-index: 40;
```
WinstonMain: Is there a z-index management strategy for React-ui?

packages/react-ui-core/src/Gmap/FreeDrawLayer.js
```diff
@@ -0,0 +1,92 @@
+import React, { PureComponent } from 'react'
+import PropTypes from 'prop-types'
+import autobind from 'autobind-decorator'
+import themed from 'react-themed'
+
+@themed(['FreeDraw', 'FreeDraw_ToolTip'], { pure: true })
+export default class FreeDrawLayer extends PureComponent {
+  static propTypes ={
+    map: PropTypes.object,
```
grenti: Should we check if the `map` prop is null before retrieving its dom in `map.getDiv()` below in multiple places? Or perhaps give the map prop a default value if needed?

stories/theme/core/Gmap.css
tadjohnston: There isn't and this is something I've suggested adding to other repos like ag.js in the past.  We would want something similar to this approach: https://css-tricks.com/handling-z-index/

packages/react-ui-core/src/Gmap/FreeDrawLayer.js
bethstevic: This is just a temporary thing until the tooltip is built, then there will be a default tooltip/button. Just not sure what it will look like once finished so not sure it makes sense to set that up right now

stories/theme/core/Gmap.css
bethstevic: I agree, we should have something set up, honestly, do get this I moved the z-index up in the browser console until it showed up. But, this is just a story so it won't affect anything ag.js side

barelyfitz: Yes, I love that idea of a z-index management system!

packages/react-ui-core/src/Gmap/FreeDrawLayer.js
bethstevic: I agree I can change that

packages/react-ui-core/src/Gmap/FreeDrawLayer.js
```diff
@@ -0,0 +1,93 @@
+import React, { PureComponent } from 'react'
+import PropTypes from 'prop-types'
+import autobind from 'autobind-decorator'
+import themed from 'react-themed'
+import { EVENTS } from './utils/const'
+import { setupEvents, removeEvent } from './utils/mapEventHelpers'
+
+@themed(['Gmap_FreeDrawTootip'], { pure: true })
+export default class FreeDrawLayer extends PureComponent {
+  static propTypes ={
```
barelyfitz: Missing space `={`

bethstevic: Thanks! missed that

packages/react-ui-core/src/Gmap/FreeDrawLayer.js
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  enableMapControls() {
+    const { map } = this.props
+    map.setOptions({
+      draggable: true,
+      zoomControl: true,
+      scrollwheel: true,
+      disableDoubleClickZoom: true,
+    })
```
barelyfitz: What happens if the map had `zoomControl:false` to start out with? Then after user completes freedraw, `zoomControl` will be turned on.

Perhaps before you disable those options, you could save the current state in `this.previousMapOptions` then restore the same values when freedraw is complete.

bethstevic: Good point :+1:

packages/react-ui-core/src/Gmap/FreeDrawLayer.js
```diff
@@ -0,0 +1,93 @@
...
+
+  @autobind
+  enableMapDraw() {
+    const { map } = this.props
+    this.disableMapControls()
+    setupEvents(map, EVENTS, {
+      onMouseDown: () => {
+        this.drawFreeHand()
+      },
+    }, true)
```
barelyfitz: Does this `onMouseDown` event need to get removed at some point? What happens if you enable map draw several times, will you end up with multiple event handlers on the map?

bethstevic: So we modified the `setupEvents` function to take a bool as the fourth argument, if the bool is true it will use `addListenerOnce` if false or not existent it will use `addListener`. That way we can just pass in `true` as the last argument and it will only set the listener once

packages/react-ui-core/src/Gmap/FreeDrawLayer.js
```diff
@@ -0,0 +1,152 @@
...
+    map.setOptions({
+      draggable: true,
+      zoomControl: true,
+      scrollwheel: true,
+      disableDoubleClickZoom: true,
+      ...mapControls,
+    })
+  }
+
+  disableMapControls() {
```
nray621: Just a heads up for when we implement this into ag.js, we will need to make sure to turn off our custom zoom event listeners as well, because we have searches firing on double clicks, scrollwheels etc independent of Google Map's zoom listeners. If we forget to turn them off, we'll potentially fire searches on those events even though the map is no longer listening for them.

packages/react-ui-core/src/Gmap/utils/mapEventHelpers.js
```diff
@@ -1,13 +1,19 @@
 export const handleEvent = (instance, name, props) =>
   event => props[name](event, props, instance)
 
-export const setupEvents = (instance, events, props) => {
+export const setupEvents = (instance, events, props, once) => {
   const eventNames = Object.keys(events)
-
+  const type = once ? 'addListenerOnce' : 'addListener'
+  const kvp = {}
```
nray621: It looks like you had this function return an object of `MapsEventListener` s in order to have easy access to to them to be removed later, but, could we name this something better than `kvp`?

packages/react-ui-core/src/Gmap/FreeDrawLayer.js
```diff
@@ -0,0 +1,152 @@
...
+    }
+    const listener = setupEvents(map, EVENTS, {
+      onMouseMove: e => {
+        poly.getPath().push(e.latLng)
+      },
+    }, false)
+
+    setupEvents(map, EVENTS, {
+      onMouseUp: () => {
+        removeEvent(listener.onMouseMove)
```
nray621: Do we need this anymore since you're setting up the event as a one time listener now?

nray621: That would mean you can remove the return value and kvp object from `setUpEvents` as well, right?

packages/react-ui-core/src/Gmap/Gmap.js
```diff
@@ -46,6 +46,9 @@ const MAP_CONTROLS = {
   mapTypeControl: true,
   scaleControl: true,
   rotateControl: true,
+  draggable: true,
```
nray621: Are these `Gmap` changes necessary? I'm guessing they're true by default, since they were on before without this code.

packages/react-ui-core/src/Gmap/FreeDrawLayer.js
bethstevic: No this one wasn't a one time listener, this is on mouse move, so we need to set up the listener and keep it on until the `mousedown` event in order to track the shape path

packages/react-ui-core/src/Gmap/Gmap.js
bethstevic: yes, we needed to add them to gmap to be able to change them for free draw

yasabere: Yup

packages/react-ui-core/src/Gmap/FreeDrawLayer.js
```diff
@@ -0,0 +1,230 @@
...
+    const { polygons } = this.state
+
+    const rawPolygon = this.addShapeToGMaps(path)
+
+    this.setState({
+      polygons: [...polygons, rawPolygon],
+    }, () => {
+      if (onMapDrawEnd) {
+        onMapDrawEnd(this.state.polygons.reduce((acc, polygon, i) => {
+          acc[i] = this.getPolygonCoords(polygon)
```
bethstevic: We shouldn't be modifying the prop `acc`, we should be avoiding mutation. Can we find a different way to write this?

packages/react-ui-core/src/Gmap/FreeDrawLayer.js
```diff
@@ -0,0 +1,230 @@
...
+    for (let i = 0; i < polygons.length; i += 1) {
+      this.clearShape(i)
+    }
+  }
+
+  formatLongLatToGmapsCoordinates(shapes) {
+    return Object.values(shapes)
+      .reduce((polygonAcc, polygon) => {
+        polygonAcc.push(polygon.reduce((acc2, point, j) => {
+          acc2[j] = { lat: point[1], lng: point[0] }
```
bethstevic: Same issue as above with the mutation

packages/react-ui-core/src/Gmap/FreeDrawLayer.js
```diff
@@ -0,0 +1,230 @@
...
+    })
+  }
+
+  @autobind
+  clearAllShapes() {
+    const { polygons } = this.state
+
+    for (let i = 0; i < polygons.length; i += 1) {
+      this.clearShape(i)
+    }
```
bethstevic: Could we just set `this.state.polygons` to an empty array?

packages/react-ui-core/src/Gmap/FreeDrawLayer.js
```diff
@@ -0,0 +1,230 @@
...
+      }
+    })
+  }
+
+  @autobind
+  clearShape(id) {
+    const polygons = [...this.state.polygons]
+    polygons[id].setMap(null)
+    polygons.splice(id, 1)
+    this.setState({
```
bethstevic: Maybe we should rename this as clearSingleShape. Although, I don't think we need this function since we might never even need this functionality. If product does decide to add this in the future we can add this function

packages/react-ui-core/src/Gmap/FreeDrawLayer.js
yasabere: Clear shape removes the shape from gmaps. It cant just be done by replacing the the array. You need to use the polygon reference to remove it

packages/react-ui-core/src/Gmap/FreeDrawLayer.js
yasabere: We need this function to call polygons[id].setMap(null)

packages/react-ui-core/src/Gmap/FreeDrawLayer.js
```diff
@@ -0,0 +1,222 @@
...
+    const shape = []
+
+    for (let i = 0; i < len; i += 1) {
+      shape.push(
+        poly.getPath()
+          .getAt(i)
+          .toUrlValue(5)
+          .split(',')
+          .map(Number)
+          .reverse()
```
sbrauer: Instead of this string manipulation you could do...
```
shape.push([poly.getPath().getAt(i).lng(), poly.getPath().getAt(i).lat()])
```
Also you might consider calling `getPath` just once before the loop and assigning it to a local.

sbrauer: I think what I proposed is more direct/clear (and probably a little faster fwiw).

grenti: I agree - I took a look at the Google Maps API and the polygon object does return the lat() lng() methods so it's easier to manipulate.

packages/react-ui-core/src/Gmap/FreeDrawLayer.js
grenti: The `acc` param is the accumulated value from the reduce function - so it's not the prop though...

Can we store the path lat long as such in the state => this.state.polygon = [[lng, lat], [lng, lat], ...]? That way if Polygon objects need to be created the FreeDrawLayer component or the GMap component will then use those path values to create distinct Polygon Google Map objects when needed. Storing the actual external api in state then forces us to have difficult means of using the data that we need (just the path values). We wouldn't need to reduce them, we'd just map over them and create Polygon objects when need be. I could be wrong here since I'm not as familiar with the Polygon Maps api so let me know if it's needed to store the entire object in React's state.

packages/react-ui-core/src/Gmap/FreeDrawLayer.js
```diff
@@ -0,0 +1,218 @@
...
+      polygons[i].setMap(null)
+    }
+    this.setState({
+      polygons: {},
+    })
+  }
+
+  formatLongLatToGmapsCoordinates(shapes) {
+    return Object.values(shapes)
+      .reduce((polygonAcc, polygon) => {
```
grenti: This seems complicated to reason about - we are reducing inside a reducer - can we simplify the data we are storing  or just break this up into smaller functions that are explicit in what they are doing?

yasabere: Ill look into this

grenti: We discussed with Orion and he proposed some really nice functional things that we'd want to discuss with Tad before attempting to include it in react-ui. For now we have decided to leave the reduce within the reduce. Thanks @yasabere and @orionstein for the call - learned a lot about ramda and FP 👍

packages/react-ui-core/src/Gmap/FreeDrawLayer.js
```diff
@@ -0,0 +1,202 @@
...
+        return polygonAcc
+      }, [])
+  }
+
+  render() {
+    const { theme, disabled } = this.props
+    const { drawing } = this.state
+    // This button is temporary - will be replaced with the tool tip when built
+    return (
+      <button
```
tadjohnston: Can we use the react-ui-core Button here?

yasabere: That will be happening in another pr @bethstevic worked on https://github.com/rentpath/react-ui/pull/337

packages/react-ui-core/src/Gmap/FreeDrawLayer.js
```diff
@@ -0,0 +1,202 @@
...
+  }
+
+  addShape(path) {
+    const rawPolygon = this.addShapeToGMaps(path)
+    this.polygons = [...this.polygons, rawPolygon]
+  }
+
+  @autobind
+  clearAllShapes() {
+    for (let i = 0; i < this.polygons.length; i += 1) {
```
tadjohnston: Can you just do the following instead?
```
this.polygons.forEach(poly => poly.setMap(null) ) // or whatever you need like this.polygons[i].setMap(null)
this.polygons = []
```

yasabere: 👍

packages/react-ui-core/src/Gmap/FreeDrawLayer.js
```diff
@@ -0,0 +1,199 @@
...
+
+  @autobind
+  clearAllShapes() {
+    this.polygons.forEach(poly => poly.setMap(null))
+    this.polygons = []
+  }
+
+  formatLongLatToGmapsCoordinates(shapes) {
+    // Should be refactored later to
+    // over(pipe(allArr, allArr), point => ({lat: point[1], lng: point[0]})(Object.values(shape)
```
tadjohnston: Why can't we do this now?

yasabere: We need to add Ramda

packages/react-ui-core/src/Gmap/FreeDrawLayer.js
```diff
@@ -0,0 +1,199 @@
...
+    shapes: {},
+    dataStyle: {},
+  }
+
+  constructor(props) {
+    super(props)
+    this.state = {
+      drawing: false,
+    }
+    this.polygons = (Object.keys(this.props.shapes).length > 0) ? [this.addShapeToGMaps(
```
tadjohnston: I think you can just do `this.polygons = Object.keys(this.props.shapes).length ? [this.addShapeToGMaps(`

tadjohnston: Anyway possible to clean this up a little? It's a bit difficult to follow.

packages/react-ui-core/src/Gmap/FreeDrawLayer.js
```diff
@@ -0,0 +1,199 @@
...
+    const len = poly.getPath().getLength()
+    const shape = []
+
+    for (let i = 0; i < len; i += 1) {
+      shape.push([poly.getPath().getAt(i).lng(), poly.getPath().getAt(i).lat()])
+    }
+    return shape
+  }
+
+  polygons = []
```
tadjohnston: Why isn't this in the constructor?

packages/react-ui-core/src/Gmap/FreeDrawLayer.js
```diff
@@ -0,0 +1,199 @@
...
+  clearAllShapes() {
+    this.polygons.forEach(poly => poly.setMap(null))
+    this.polygons = []
+  }
+
+  formatLongLatToGmapsCoordinates(shapes) {
+    // Should be refactored later to
+    // over(pipe(allArr, allArr), point => ({lat: point[1], lng: point[0]})(Object.values(shape)
+
+    return Object.values(shapes)
```
tadjohnston: Can you explain this one? Using a nested reduce seems convoluted.  What's it for?

packages/react-ui-core/src/Gmap/FreeDrawLayer.js
```diff
@@ -0,0 +1,199 @@
...
+      )
+    }
+  }
+
+  getPolygonCoords(poly) {
+    const len = poly.getPath().getLength()
+    const shape = []
+
+    for (let i = 0; i < len; i += 1) {
+      shape.push([poly.getPath().getAt(i).lng(), poly.getPath().getAt(i).lat()])
```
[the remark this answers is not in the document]
yasabere: 👍

packages/react-ui-core/src/Gmap/FreeDrawLayer.js
yasabere: Fixed

packages/react-ui-core/src/Gmap/__tests__/FreeDrawLayer-test.js
```diff
@@ -0,0 +1,222 @@
...
+}
+const lat = jest.fn()
+const lng = jest.fn()
+const shape = [
+  { lat: 33.90871, lng: -84.56927 },
+  { lat: 33.91783, lng: -84.6297 },
+  { lat: 33.92239, lng: -84.64618 },
+]
+const expectedPoints = [[-84.56927, 33.90871], [-84.6297, 33.91783], [-84.64618, 33.92239]]
+getPath.mockReturnValue({
```
tadjohnston: Can't you just combine this onto the `jest.fn()`???

packages/react-ui-core/src/Gmap/FreeDrawLayer.js
```diff
@@ -0,0 +1,199 @@
...
+    this.polygons.forEach(poly => poly.setMap(null))
+    this.polygons = []
+  }
+
+  formatLongLatToGmapsCoordinates(shapes) {
+    // Should be refactored later to
+    // over(pipe(allArr, allArr), point => ({lat: point[1], lng: point[0]})(Object.values(shape)
+
+    return Object.values(shapes)
+      .reduce((polygonAcc, polygon) => {
```
grenti: Can we map/reduce this? We can map over the properties in the shape object then just reduce all the arrays within that property to create a shape based on it's arrays

packages/react-ui-core/src/Gmap/FreeDrawLayer.js
grenti: @orionstein suggested it on Friday and he wanted us to talk to you about installing ramda and using it. I think he wanted to have a library with all the lenses/over/pipe ramda functions we'd need before installing it though.

packages/react-ui-core/src/Gmap/FreeDrawLayer.js
yasabere: Done

packages/react-ui-core/src/Gmap/FreeDrawLayer.js
yasabere: Fixed

packages/react-ui-core/src/Gmap/FreeDrawLayer.js
```diff
@@ -0,0 +1,202 @@
...
+    this.setState({
+      drawing: false,
+    })
+  }
+
+  disableMapControls() {
+    const { map, mapControls } = this.props
+    map.setOptions({
+      ...MAP_CONTROLS_DISABLED,
+      ...mapControls,
```
tadjohnston: That doesn't work... What if someone passed in `zoomControl: true` in the mapControls prop???

packages/react-ui-core/src/Gmap/FreeDrawLayer.js
```diff
@@ -0,0 +1,202 @@
...
+  @autobind
+  enableMapDraw() {
+    const { map, multipleShapes } = this.props
+    this.disableMapControls()
+
+    if (!multipleShapes) {
+      this.clearAllShapes()
+    }
+    this.events = setupEvents(map, EVENTS, {
+      onMouseDown: () => {
```
tadjohnston: Can't this be better written:
```
onMouseDown: this.drawFreeHand
```

You remove the unnecessary anonymous function.

packages/react-ui-core/src/Gmap/FreeDrawLayer.js
tadjohnston: Yeah, completely 💯 on ramda.

packages/react-ui-core/src/Gmap/Gmap.js
tadjohnston: Not getting why we need them either..

packages/react-ui-core/src/Gmap/FreeDrawLayer.js
yasabere: 👍

packages/react-ui-core/src/Gmap/FreeDrawLayer.js
```diff
@@ -0,0 +1,193 @@
...
+    })
+    this.setState({
+      drawing: false,
+    })
+  }
+
+  disableMapControls() {
+    const { map } = this.props
+    map.setOptions({
+      ...enableControls(false),
```
tadjohnston: Why do you need to create another object here?

packages/react-ui-core/src/Gmap/utils/const.js
```diff
@@ -0,0 +1,25 @@
+export const EVENTS = {
```
tadjohnston: Did you mean to remove this from somewhere else? Also, the name `const.js` or `EVENTS` doesn't make sense here. What are these events in reference to? The map? The markers? All of them? Thats what I would assume.

packages/react-ui-core/src/Gmap/utils/mapEventHelpers.js
```diff
@@ -1,13 +1,19 @@
...
     if (props[name] && typeof props[name] === 'function') {
       const event = handleEvent(instance, name, props)
-      window.google.maps.event.addListener(instance, events[name], event)
+      kvp[name] = window.google.maps.event[type](instance, events[name], event)
     }
   })
+  return kvp
+}
+
+export const removeEvent = event => {
```
tadjohnston: Can you add a test for this?

packages/react-ui-core/src/Gmap/utils/mapEventHelpers.js
tadjohnston: Can you also add a test for this?

packages/react-ui-core/src/Gmap/FreeDrawLayer.js
```diff
@@ -0,0 +1,193 @@
...
+    shapes: {},
+    dataStyle: {},
+  }
+
+  constructor(props) {
+    super(props)
+    this.state = {
+      drawing: false,
+    }
+    this.polygons = Object.keys(this.props.shapes).length ? [this.addShapeToGMaps(
```
grenti: Can we extract `Object.keys(this.props.shapes).length` into a method like this: 
```
  hasShape() {
    return Object.keys(this.props.shapes).length
  }
```

grenti: Actually now that I think of it, can we put this into an initializePolygons() method? 

```
  initializePolygons() {
    if (this.hasShape()) {
      const transposedCoordinates = this.formatLongLatToGmapsCoordinates(this.props.shapes)
      return this.addShapeToGMaps(transposedCoordinates)
    }

    return []
  }
```

Also I think both `addShapeToGMaps` and `formatLongLatToGmapsCoordinates` should just return whatever they manipulated instead of trying to pick the first index. If we intend to only show one polygon then we can control that in ag.js and pass only one shape's data. I think this component should just be concerned about giving the right information to GMaps for shape creation and deletion.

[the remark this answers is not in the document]
yasabere: The only time the google maps deals with this format is when taking in new paths. Never when returning paths so it doesn't make sense to accept it

yasabere: This is refactored now

packages/react-ui-core/test/google-maps-mock.js
```diff
@@ -0,0 +1,1055 @@
+// Found at: https://github.com/prasathmani/google-maps-mock-api
```
tadjohnston: We'll need to clean this up and either decide to use this file or the logic we currently have in `test/shim.js`.

packages/react-ui-core/src/Gmap/FreeDrawLayer.js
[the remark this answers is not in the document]
tadjohnston: I agree with you though @grenti , I don't think we should be manipulating data coming in. It should be in the format Gmaps wants it in so I/O is the same.  It may not be the same data if rent-js used it and then we'd have to rewrite it. We should probably strip it out TBH.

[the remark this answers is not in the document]
tadjohnston: @yasabere it actually makes sense what @grenti is saying. You're assuming that the format is always in at listing service format and are then mutating it into what Gmaps wants.  That seems wrong.

packages/react-ui-core/src/Gmap/utils/mapEventHelpers.js
```diff
@@ -1,13 +1,19 @@
 export const handleEvent = (instance, name, props) =>
   event => props[name](event, props, instance)
 
-export const setupEvents = (instance, events, props) => {
+export const setupEvents = (instance, events, props, once) => {
   const eventNames = Object.keys(events)
-
+  const type = once ? 'addListenerOnce' : 'addListener'
+  const eventFunctions = {}
```
tadjohnston: How about `eventHandlers` instead?

packages/react-ui-core/src/Gmap/FreeDrawLayer.js
```diff
@@ -0,0 +1,176 @@
...
+import { setupEvents, removeEvent } from './utils/mapEventHelpers'
+
+const enableControls = (enable = true) => ({
+  draggable: enable,
+  zoomControl: enable,
+  scrollwheel: enable,
+  disableDoubleClickZoom: enable,
+})
+
+@themed(['Gmap_FreeDrawTootip'], { pure: true })
```
tadjohnston: This name should be the same as the component.

yasabere: There is a separate pr for this tool top

packages/react-ui-core/src/Gmap/utils/mapEventHelpers.js
yasabere: 👍

PR #335, opened 2018-06-21
feat(ratings): adds support for top ratings summary

dgavigan:
Adds new prop of `ratingValue` to display as a key legend

screen shot 2018-06-21 at 3 49 48 pm

[Card - Reviews - Ratings top ratings component (RDS) ](https://rentpath.leankit.com/card/682447309)

### DIFF
--- a/packages/react-ui-core/src/Ratings/RatingBar.js
+++ b/packages/react-ui-core/src/Ratings/RatingBar.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types'
 import themed from 'react-themed'
 import classnames from 'classnames'
 
-@themed(/^(RatingBar|Label)/, {
+@themed(/^(RatingBar|Label|Legend)/, {
   pure: true,
 })
 
@@ -19,6 +19,21 @@ export default class RatingBar extends PureComponent {
     theme: {},
     score: 0,
     maxScore: 5,
+  }
+
+  get ratingLegend(){
+    const {
+      theme, 
+      score,
+      ratingValue
+    } = this.props
+
+    return (
+      <div className={theme.Legend}>
+        <div className={theme.RatingBar_RatingValue}> {ratingValue} </div>
+        <div className={theme.RatingBar_LegendIcon}></div>
+      </div>
+    )
   }
 
   get ratingItems() {
@@ -43,7 +58,7 @@ export default class RatingBar extends PureComponent {
       </div>
     )
   }
-
+  
   render() {
     const {
       theme,
@@ -51,6 +66,7 @@ export default class RatingBar extends PureComponent {
       className,
       score,
       maxScore,
+      ratingValue,
       ...props
     } = this.props
 
@@ -64,8 +80,14 @@ export default class RatingBar extends PureComponent {
         }
         {...props}
       >
+        {ratingValue && this.ratingLegend}
         {this.ratingItems}
-        {label &&
+        {ratingValue && 
+          <div className={theme.Label}>
+            {score}
+          </div>
+        }
+        {label && 
           <div className={theme.Label}>
             {label}
           </div>
--- a/stories/core/RatingBar.js
+++ b/stories/core/RatingBar.js
@@ -57,3 +57,61 @@ export const HandleRatingClick = (
     label="Click a Star"
   />
 )
+
+export const RatingSummary = (
+  <RatingBar
+    className={theme.themedRating_summary}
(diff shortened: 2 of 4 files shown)


packages/react-ui-core/src/Ratings/RatingBar.js
```diff
@@ -21,6 +21,21 @@ export default class RatingBar extends PureComponent {
     maxScore: 5,
   }
 
+  get ratingLegend(){
```
barelyfitz: Be sure to run `yarn lint` before you commit, also see if you can set up your editor to display eslint errors so you can catch them sooner. There are numerous lint violations (missing or extra spacing, etc)

packages/react-ui-core/src/Ratings/RatingBar.js
```diff
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types'
 import themed from 'react-themed'
 import classnames from 'classnames'
 
-@themed(/^(RatingBar|Label)/, {
+@themed(/^(RatingBar|Label|Legend)/, {
```
barelyfitz: I'm not sure why react-ui uses `Label` here, because generally we try to keep each component with unique classnames like `RatingBar_Label` (perhaps they are using `Label` as a standard class across many of the components.

The combination of Label and Legend is a little ambiguous to me. Maybe `RatingBar_Prefix` would be a better name for that class.

packages/react-ui-core/src/Ratings/RatingBar.js
```diff
@@ -43,14 +58,15 @@ export default class RatingBar extends PureComponent {
...
-
+  
   render() {
     const {
       theme,
       label,
       className,
       score,
       maxScore,
+      ratingValue,
```
barelyfitz: All these props could be confusing (what is `ratingValue`, what is `score`, they seem to imply the same thing). Maybe call this `prefixLabel` or something like that.

packages/react-ui-core/src/Ratings/RatingBar.js
```diff
@@ -21,6 +21,21 @@ export default class RatingBar extends PureComponent {
...
+      theme, 
+      score,
+      ratingValue
+    } = this.props
+
+    return (
+      <div className={theme.Legend}>
+        <div className={theme.RatingBar_RatingValue}> {ratingValue} </div>
+        <div className={theme.RatingBar_LegendIcon}></div>
+      </div>
```
barelyfitz: This assumes that the developer will want to display an icon. How about somthing like this, and let the developer pass in whatever they want?
```
<div className={theme.RatingBar_Prefix}>{this.props.prefixLabel}</div>
```
Then if I want an icon  I can just pass in whatever content I want.

packages/react-ui-core/src/Ratings/RatingBar.js
```diff
@@ -64,8 +80,14 @@ export default class RatingBar extends PureComponent {
...
       >
+        {ratingValue && this.ratingLegend}
         {this.ratingItems}
-        {label &&
+        {ratingValue && 
+          <div className={theme.Label}>
+            {score}
+          </div>
+        }
+        {label &&
```
barelyfitz: Can I just pass in "18 of 30" as my label, instead of coding this special case to meet one particular use? Lets try to keep this component generic so developers can use it in a variety of circumstances.

barelyfitz:
For the description of this PR - move the card link to the first line, since this is the standard and makes it easy to find. I also like to prefix it with `# ` or `## ` to make it a heading and increase the font size.

packages/react-ui-core/src/Ratings/RatingBar.js
```diff
@@ -21,6 +21,21 @@ export default class RatingBar extends PureComponent {
...
+  get ratingLegend(){
+    const {
+      theme, 
+      score,
+      ratingValue
+    } = this.props
+
+    return (
+      <div className={theme.Legend}>
+        <div className={theme.RatingBar_RatingValue}> {ratingValue} </div>
```
tadjohnston: Please remove the spaces before and after `{ratingValue}`.

packages/react-ui-core/src/Ratings/RatingBar.js
dgavigan: @barelyfitz will do

dgavigan:
thanks for the notes gentlemen ... will fix and create the PR in rent-js

PR #338, opened 2018-07-03
Remove slideToIndex

justinrex:
[Card](https://rentpath.leankit.com/card/689327821)

This slideToIndex was causing sliding to index 0 when unexpected.
Turns out it was old code from visual search that is no longer needed.


aleima11:
Just a side question, has anyone removed all of the other visual search logic? I highly doubt they did from AMP

nray621:
@aleima11 we still have visual search, this specifically was just some zombie code for a previous implementation of visual search for the BLC3 variation

PR #344, opened 2018-07-09
fix(FreeDrawLayer): add onEnableMapDraw

yasabere:

Add onEnableMapDraw to freeDrawLayer

Allows for onEnableMapDraw call back to be fired when enabling draw mode in free draw

### DIFF
--- a/packages/react-ui-core/src/Gmap/FreeDrawLayer.js
+++ b/packages/react-ui-core/src/Gmap/FreeDrawLayer.js
@@ -21,6 +21,7 @@ export default class FreeDrawLayer extends PureComponent {
     theme: PropTypes.object,
     map: PropTypes.object,
     onMapDrawStart: PropTypes.func,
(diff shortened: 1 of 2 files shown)


tadjohnston:
Explanation?

packages/react-ui-core/src/Gmap/FreeDrawLayer.js
```diff
@@ -81,12 +82,13 @@ export default class FreeDrawLayer extends PureComponent {
 
   @autobind
   enableMapDraw() {
```
grenti: can you rename this `enableMapDraw` method to `handleShapeDrawing`? I just noticed that it seemed like the same as the prop and it's slightly confusing.

hernquist: To Anye's point, it is potentially confusing.

tadjohnston:
I don't think we need this anymore do we?

PR #348, opened 2018-07-11
fix(react-ui-ag): pass propertyName props through, prevent passing them to ApartmentComplex

mneumark:

* pass propertyName props to `ListingComponents.PropertyName`
* prevent propertyName props from being passed down to `Listing`
* also added support for `address` props in the same way while I was at it

Note: I primarily did this because it was causing warnings in `ag.js`. I am trying to prevent warnings from printing while in tests in `ag.js`.


WinstonMain:
👍

PR #350, opened 2018-07-11
refactor(freedraw): refactor freedraw

tadjohnston:

Freedraw is too coupled and needs to be broken up. Some of the components are redundant from what we
already have.

- decoupled the layer from the other components in there
- removed the banner completely, it's just a text field
- removed the button, you can just use the ToggleButton instead since
that's what it really is
- added an `enabled` prop to the FreeDrawLayer that drives the usability
of the drawing


package.json
```diff
@@ -51,6 +51,7 @@
     "@types/node": "^10.3.4",
     "@types/react": "^16.4.0",
     "@types/storybook__react": "^3.0.7",
+    "autobind-decorator": "^2.1.0",
```
bethstevic: How did we not have autobind before?

tadjohnston: This is in the root level so StoryBook can use it.

packages/react-ui-core/src/Gmap/FreeDrawLayer.js
```diff
@@ -1,12 +1,8 @@
-import React, { PureComponent } from 'react'
+import { PureComponent } from 'react'
```
bethstevic: Don't we need react since this is a component?

tadjohnston: Not unless you return jsx 😄

PR #355, opened 2018-07-18
feat(core): adds Picture tag to Photo component

ninomichaels:

### DIFF
--- a/packages/react-ui-core/src/Photo/Photo.js
+++ b/packages/react-ui-core/src/Photo/Photo.js
@@ -4,6 +4,12 @@ import classnames from 'classnames'
 import autobind from 'autobind-decorator'
 import themed from 'react-themed'
 
+const pictureSourcesPropTypes = PropTypes.shape({
+  type: PropTypes.string,
+  media: PropTypes.string,
+  srcset: PropTypes.string.isRequired,
+})
+
 @themed(['Photo'], {
   pure: true,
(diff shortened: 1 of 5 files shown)


packages/react-ui-core/src/Photo/Photo.js
```diff
@@ -75,23 +106,31 @@ export default class Photo extends PureComponent {
...
+      data-tid="photo"
+      className={classnames(
+        className,
+        theme.Photo
+      )}
+      onError={this.fallback}
+      {...rest}
+    />)
+
+    if (error || !this.sourceTags) return imageTag
```
lfender6445: we want to return an imageTag if there is an error?

lfender6445:
im 👍 other than comment

packages/react-ui-core/src/Photo/Photo.js
ninomichaels: this ensures the fallback (or nothing) is displayed, otherwise it will try to select a "better" source from the `srcset`, even if it doesn't exist, and you end up with this:

screen shot 2018-07-20 at 8 59 50 am

ninomichaels: Since we have no fallback on Rent at this time, the desired outcome is: 

screen shot 2018-07-20 at 9 02 55 am

PR #356, opened 2018-07-20
refactor(react-ui-core): refactor events in free draw

nray621:

Rather than set mouseMove event listeners on click, have their functions only draw when the mouse is actually down. 
onMouseDown sets mouseDown flag to true
onMouseUp: 
1.  sets mouseDown flag to false
2. If there is a valid polygon handles the drawing end and removes all even listeners


packages/react-ui-core/src/Gmap/FreeDrawLayer.js
```diff
@@ -125,13 +131,20 @@ export default class FreeDrawLayer extends PureComponent {
 
     this.events = {
       ...setupEvents(map, GMAP_EVENTS, {
-        onMouseMove: e => this.polyline.getPath().push(e.latLng),
-        onTouchStart: e => this.polyline.getPath().push(e.latLng),
+        onMouseDown: () => { this.mouseDown = true },
+        onTouchStart: () => { this.mouseDown = true },
+        onMouseMove: e => this.handleMouseMove(e),
+        onTouchMove: e => this.handleMouseMove(e),
```
tadjohnston: This should be:

onTouchMove: this.handleMouseMove,

packages/react-ui-core/src/Gmap/FreeDrawLayer.js
```diff
@@ -125,13 +131,20 @@ export default class FreeDrawLayer extends PureComponent {
 
     this.events = {
       ...setupEvents(map, GMAP_EVENTS, {
-        onMouseMove: e => this.polyline.getPath().push(e.latLng),
-        onTouchStart: e => this.polyline.getPath().push(e.latLng),
+        onMouseDown: () => { this.mouseDown = true },
+        onTouchStart: () => { this.mouseDown = true },
```
tadjohnston: I’d prefer these two events call one function that sets this value.

packages/react-ui-core/src/Gmap/FreeDrawLayer.js
nray621: 👍

packages/react-ui-core/src/Gmap/FreeDrawLayer.js
nray621: sure, I'll change that

PR #363, opened 2018-07-27
Remove schema

cwklausing:
Taking @tadjohnston 's earlier branch, updating it to remove/ignore libs, remove remainder of schema references.

This compliments the work happening in [this PR](https://github.com/rentpath/ag.js/pull/4768), and [this Card](https://rentpath.leankit.com/card/692916862)


.gitignore
```diff
@@ -1,5 +1,5 @@
-lib
```
tadjohnston: Can't we just revert this file to origin/master?

colinrymer: Agreed with tad; this doesn't seem to be something we'd want to commit.

PR #365, opened 2018-07-31
Cloudinary updates

aleima11:
[card](https://trello.com/c/fSH5wvm1/2-cloudinary-on-ag)
- The URL being constructed here no longer works
- Passing in the URL now instead

### DIFF
--- a/packages/react-ui-ag/src/Listings/Listing.js
+++ b/packages/react-ui-ag/src/Listings/Listing.js
@@ -199,9 +199,9 @@ export default class Listing extends Component {
     }
 
     if (!isActive) {
-      const { server, dimensions } = photos
+      const { server } = photos
       const firstPhoto = this.listingPhotos[0] || {}
-      listing.photo = { url: `${server}${firstPhoto.path}${dimensions}` }
+      listing.photo = { url: `${server}${firstPhoto.path}` }
     }
 
     return (
--- a/stories/ag/Listings/DesktopListing.js
+++ b/stories/ag/Listings/DesktopListing.js
@@ -14,8 +14,7 @@ const props = {
     className: 'favorite-button',
   },
   photos: {
-    server: 'https://image.rent.com/',
-    dimensions: '280-120',
+    server: 'https://rentpath-res.cloudinary.com/',
     showIndex: true,
   },
   ratings: {
--- a/packages/react-ui-ag/src/Listings/MobileMapListing.js
+++ b/packages/react-ui-ag/src/Listings/MobileMapListing.js
@@ -182,9 +182,9 @@ export default class MobileMapListing extends Component {
     }
 
     if (!isActive) {
-      const { server, dimensions } = photos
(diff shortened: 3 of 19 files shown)


packages/react-ui-ag/src/Listings/Listing.js
```diff
@@ -31,7 +31,12 @@ export default class Listing extends Component {
     theme: PropTypes.object,
     className: PropTypes.string,
     onClick: PropTypes.func,
-    photos: PropTypes.object,
+    photos: PropTypes.arrayOf(
```
tadjohnston: This isn't what you think it is.. https://github.com/rentpath/react-ui/pull/365/files#diff-be83dd49bf57212006170863d5504aa5L86

We should probably just rename this prop to `photoConfig` or something else.

aleima11: yeah makes more sense. Doesn't make sense anymore to construct the url in react-ui.

aleima11: And I went redo crazy before I realized. This is super WIP.

tadjohnston:
That should be the responsibility of the listing then and none of it should be done here.  The reason that configuration was done was to allow for multiple servers (load balancers) to be used and multiple dimensions for srcset usage. It was probably overkill to ever have react-ui even worry about it in the first place.

aleima11:
So is there an issue with me removing that logic from react-ui and just having the photo url passed in?

tadjohnston:
Nope, no issue, but then we can remove the `photo` prop from Listing https://github.com/rentpath/react-ui/pull/365/files#diff-cbc7d33b278cde12ded8005102f3bf1eR34.

aleima11:
yeah for sure, those changes are all wrong.

aleima11:
I left the server and then path, which will be the whole second part of the url. This way you can still do srcset. Just have to build in dimensions to the path before hand.

stories/ag/Listings/DesktopListing.js
```diff
@@ -15,7 +15,6 @@ const props = {
   },
   photos: {
     server: 'https://image.rent.com/',
-    dimensions: '280-120',
```
nray621: Sorry, I know this is a nitpick and a bunch of stories are currently broken anyway, but would you mind adding the dimensions to the URLs in `stories/ag/Listings/baseListing.js` so the story still has the proper dimensions?

packages/react-ui-core/package.json
```diff
@@ -53,6 +53,7 @@
     "classnames": "^2.2.6",
     "form-serialize": "0.7.2",
     "prop-types": "^15.6.1",
+    "react": "^16.4.1",
```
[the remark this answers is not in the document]
aleima11: oopsies. Will do

tadjohnston:
Please make sure you use `git cz` format on your commit before merge.

PR #369, opened 2018-08-08
ULA Text Refactor

LinnJS:
[Card](https://rentpath.leankit.com/card/682372371)

Acceptance Criteria:

Update ULA and Last Updated text logic to change to as follows
If there is ULA then show only ULA
IF there is ULA and Last Updated information show only ULA
If there is no ULA and there is last updated information then show the last updated information
If there is no ULA and No Last updated information then show a blank row
In all situations remove the "clock" icon currenlty shown between ULA and last updated.

How it related to `react-ui`:
This card asks not to display `ULA`(UnitLevelAvailability: "9 units available") and `LastUpdated` at the same time and primarily display `ULA` and then fall back to LastUpdated when `ULA` not present.

Made changes in react-ui to remove not used `div` to match props currently in `DesktopListing.js`

Link to related PR:
https://github.com/rentpath/ag.js/pull/4853

What it will look like:
screen shot 2018-08-10 at 12 31 50 pm

### DIFF
--- a/packages/react-ui-ag/src/Listings/DesktopListing.js
+++ b/packages/react-ui-ag/src/Listings/DesktopListing.js
@@ -9,7 +9,7 @@ import Listing from './Listing'
   'DesktopListing',
   'BedsAndBaths',
   'UnitLevelAvailabilityAndLastUpdated',
-  'LastUpdated',
+  'UlaText',
   'Phone',
 ], { pure: true })
 export default class DesktopListing extends PureComponent {
@@ -47,7 +47,7 @@ export default class DesktopListing extends PureComponent {
           <ListingComponents.Availability /> :
           (
             <div className={theme.UnitLevelAvailabilityAndLastUpdated}>
-              {ulaText && <div className={theme.UlaText}>{ulaText}</div>}
+              {ulaText && <div className={theme.UlaText} data-tid="ulaText">{ulaText}</div>}
             </div>
           )
         }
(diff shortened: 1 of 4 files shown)


packages/react-ui-ag/src/Listings/DesktopListing.js
```diff
@@ -30,7 +30,7 @@ export default class DesktopListing extends PureComponent {
 
   get renderInfo() {
     const { theme, listing, ratings, propertyName } = this.props
-    const { singleFamily, rating, lastUpdated, phone } = listing
+    const { singleFamily, rating, ulaText, phone } = listing
```
LinnJS: new consolidated prop which will be comprised of a selector that returns `ula ? displayUla : DisplayLastUpdated`

stories/theme/ag/large/Listings/DesktopListing.css
```diff
@@ -216,7 +216,7 @@ a.DesktopListing .Listing_CtaButton {
   margin-top: 2px;
 }
 
-.DesktopListing .LastUpdated {
```
LinnJS: Updated CSS className to reflect naming changes

packages/react-ui-ag/src/Listings/DesktopListing.js
```diff
@@ -47,8 +47,7 @@ export default class DesktopListing extends PureComponent {
           <ListingComponents.Availability /> :
           (
             <div className={theme.UnitLevelAvailabilityAndLastUpdated}>
-              <ListingComponents.UnitLevelAvailability />
```
LinnJS: Removed ULA component since its being replaced with `ulaText`

tadjohnston: Can we remove the component then from `ListingComponents`?

tadjohnston: On second thought, just leave it. It's not hurting anything.

PR #370, opened 2018-08-09
feat(Form): core Form Tweaks

mneumark:

[Story](https://rentpath.leankit.com/card/678468241)

* moved `LeadForm` fields into `Form/SpecificFields`
* added additional typical fields into `SpecificFields` directory
* created `FormFields` export that provides access to all the fields
* added `Validated` version of all fields

I was working on a form for rent-js and the way they were doing it was cumbersome. These changes are intended as a way to simplify building forms. Let me know if you have any questions.


packages/react-ui-core/src/Form/SpecificFields/Message.js
```diff
@@ -1,12 +1,9 @@
...
 import PropTypes from 'prop-types'
 import themed from 'react-themed'
-import { Field } from '../../Form'
-
-@themed(['Label_Textarea'], {
-  pure: true,
-})
+import Field from '../Field'
 
+@themed(['Message', 'Label_Textarea'], { pure: true })
```
tadjohnston: Weird. This should have never been themed in the first place.  `Label_Textarea` isn't in this component so it's not needed.

mneumark: I'll remove it. I was just making sure i didn't break something or make it backwards incompatible.

packages/react-ui-core/src/Form/LabelHidingField.js
```diff
@@ -0,0 +1,77 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+import themed from 'react-themed'
+import autobind from 'autobind-decorator'
+import Field from './Field'
+
+@themed(/^LabelHidingField/)
+export default class LabelHidingField extends Component {
```
tadjohnston: Like the name!

packages/react-ui-core/src/Form/LabelHidingField.js
```diff
@@ -0,0 +1,77 @@
...
+  static propTypes = {
+    theme: PropTypes.object,
+    name: PropTypes.string,
+    type: PropTypes.string,
+    label: PropTypes.string,
+    value: PropTypes.string,
+    defaultValue: PropTypes.string,
+    onChange: PropTypes.func,
+    nodeType: PropTypes.any,
```
tadjohnston: I don't think `any` will work here. An object would break this. You could do:
```
    nodeType: PropTypes.oneOfType([
      PropTypes.node,
      PropTypes.func,
      PropTypes.object,
    ]),
```
and apply object and such to Field or just make it just a:
```
  nodetype: PropTypes.node
```
??

packages/react-ui-core/src/Form/RadioGroup.js
```diff
@@ -8,7 +8,7 @@ import omit from 'lodash/omit'
 import isEqual from 'lodash/isEqual'
 import RadioButton from './RadioButton'
 
-@themed('*')
+@themed(/^RadioGroup/)
```
tamars2: 👍

PR #372, opened 2018-08-10
673931956 - Update FreeDrawLayer to accommodate multiple shapes

grenti:
[Card](https://rentpath.leankit.com/card/673931956)


grenti:
Will do - we realized we didn’t use it earlier and so I was going to update the pr with tests and git after I remove WIP.

packages/react-ui-core/src/Gmap/FreeDrawLayer.js
```diff
@@ -79,6 +79,9 @@ export default class FreeDrawLayer extends PureComponent {
     const polygonCoords = this.getPolygonCoords(this.polygon)
 
     if (polygonCoords.length) this.finishDraw(polygonCoords)
+
+    this.endDraw()
```
tadjohnston: I think you need to add tests to make sure these two get fired here.

stories/core/ExampleComponents/FreeDrawExample.js
```diff
@@ -12,37 +12,49 @@ export default class FreeDrawExample extends PureComponent {
...
+      shapes: shape && { ...shapes, [shapeId]: shape },
+      enabled: this.props.allowMultipleShapes,
     })
   }
 
   @autobind
   handleClick() {
-    this.setState({ enabled: !this.state.enabled })
+    const existingShapes = this.props.shapes
+    const { enabled, shapes: _ } = this.state
```
tadjohnston: `_` usually denotes you aren't using that prop. Cant' we just rename it to something else? I'd prefer the code below to be readable.

stories/core/ExampleComponents/FreeDrawExample.js
```diff
@@ -12,37 +12,49 @@ export default class FreeDrawExample extends PureComponent {
...
+
+    const shapes = !enabled && Object.keys(_).length > 0 && _ !== existingShapes
```
tadjohnston: `Object.keys(_).length > 0` should just be `Object.keys(_).length`

tadjohnston: I also don't think we should really be using both props and state to write conditionals.  What happens if `props.shapes` changes is that going to be an issue?

grenti: I think I'm going to take out the state shape logic - I think it was intended to store the shapes and now since we are trying to reconcile both it's getting out of hand. I'll just reset the state on click back to the props passed in.

stories/core/ExampleComponents/FreeDrawExample.js
```diff
@@ -12,37 +12,45 @@ export default class FreeDrawExample extends PureComponent {
...
   @autobind
   handleOnDrawEnd(shape) {
+    const { shapes } = this.state
+
+    if (this.props.onDrawEnd) this.props.onDrawEnd(shape)
+    const shapeId = Object.keys(shapes).length
     this.setState({
-      shapes: shape && { 0: shape },
-      enabled: false,
+      shapes: shape && { ...shapes, [shapeId]: shape },
```
tadjohnston: I thought you were removing this?

grenti: I removed the logic from `handleClick` that updated the shapes in state. The way the component is written we have to use the state to keep all the shapes that have been drawn in the Multi-shape scenario. We can refactor this component to handle multi-shapes but since we're trying to get this released on Thursday I was trying to be as conservative as possible.

grenti:
Ok I'll squash - I was still making changes based on our conversation but we do need to get this merged so we can publish and use for ag.js changes going in tomorrow.

PR #375, opened 2018-08-14
fix(listing): listing theme fix

tadjohnston:

* fixed PropertyLink on listings
* added error class to Photo

### DIFF
--- a/packages/react-ui-ag/src/Listings/MobileListing.js
+++ b/packages/react-ui-ag/src/Listings/MobileListing.js
@@ -66,7 +66,7 @@ export default class MobileListing extends PureComponent {
         <ListingComponents.Price />
         <a
           href={listing.url}
-          className={theme.PropertyLink}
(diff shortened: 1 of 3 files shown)


lfender6445:
👍

packages/react-ui-ag/src/Listings/MobileListing.js
```diff
@@ -5,7 +5,7 @@ import classnames from 'classnames'
 import { ListingComponents, Text } from '@rentpath/react-ui-core'
 import Listing from './Listing'
 
-@themed(/^MobileListing/, { pure: true })
+@themed(/^(MobileListing|PropertyLink$)/, { pure: true })
```
lfender6445: why not just change theme key to theme.MobileListing_PropertyLink

tadjohnston: I guess we could do that instead. Makes more sense, I'll change it.

PR #377, opened 2018-08-16
feat(form): controlled / uncontrolled fields

tadjohnston:

* added input mask option for input types
* added hint, label and error components and clases for RadioGroup
* added controlled / uncontrolled functionality for Input and
RadioGroup
* added proper type of 'checked' to RadioButton

BREAKING CHANGES:
removed placheholders for inputs

### DIFF
--- a/packages/react-ui-core/src/Form/Email.js
+++ b/packages/react-ui-core/src/Form/Email.js
@@ -12,7 +12,6 @@ export default class Email extends PureComponent {
   static defaultProps = {
     name: 'email',
     type: 'email',
-    placeholder: 'Email Address',
   }
 
   render() {
--- a/packages/react-ui-core/src/Form/Form.js
+++ b/packages/react-ui-core/src/Form/Form.js
@@ -40,6 +40,11 @@ export default class Form extends Component {
      * Serializes form data when true.
      */
     serialize: PropTypes.bool,
+
+    /**
+     * Serialize options when serialize is on
+     */
+    serializeOptions: PropTypes.object,
   }
 
   static defaultProps = {
@@ -47,15 +52,18 @@ export default class Form extends Component {
     action: '#',
     method: 'POST',
     noValidate: true,
+    serializeOptions: {
+      hash: true,
+    },
   }
 
   onSubmit = event => {
-    const { onSubmit, serialize } = this.props
+    const { onSubmit, serialize, serializeOptions } = this.props
     event.preventDefault()
 
     if (onSubmit) {
       if (serialize) {
-        onSubmit(event, serializeForm(event.target, { hash: true }))
+        onSubmit(event, serializeForm(event.target, serializeOptions))
       } else {
         onSubmit(event)
       }
@@ -66,6 +74,7 @@ export default class Form extends Component {
     const {
       theme,
       serialize,
+      serializeOptions,
       className,
       ...props
     } = this.props
--- a/packages/react-ui-core/src/Form/Input.js
+++ b/packages/react-ui-core/src/Form/Input.js
@@ -2,6 +2,8 @@ import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 import classnames from 'classnames'
 import themed from 'react-themed'
+import InputMask from 'react-input-mask'
+import autobind from 'autobind-decorator'
 
 @themed(/^Input/, {
   pure: true,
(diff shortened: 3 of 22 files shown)


packages/react-ui-core/src/Form/Email.js
```diff
@@ -12,7 +12,6 @@ export default class Email extends PureComponent {
   static defaultProps = {
     name: 'email',
     type: 'email',
-    placeholder: 'Email Address',
```
lfender6445: i see we are omitting placeholder as a prop, does this still mean placehold is available as in attribute with ...props?

lfender6445: https://github.com/rentpath/react-ui/blob/af25ce3495924b396fa25f5b65bb7a4bea9a50e9/packages/react-ui-core/src/Form/Email.js#L19

packages/react-ui-core/src/Form/Form.js
```diff
@@ -40,22 +40,30 @@ export default class Form extends Component {
...
+    serializeOptions: PropTypes.object,
   }
 
   static defaultProps = {
     theme: {},
     action: '#',
     method: 'POST',
     noValidate: true,
+    serializeOptions: {
+      hash: true,
```
lfender6445: what does this do

packages/react-ui-core/src/Form/RadioGroup.js
```diff
@@ -108,29 +149,68 @@ export default class RadioGroup extends Component {
...
+    const props = {
+      label: null,
+      error: null,
+      hint: null,
+    }
+
     if (!this.fields.length) return null
 
+    if (label) {
+      props.label = createElement(...parseArgs(label, Label, {
```
lfender6445: is there an advantage of using createElement as opposed to straight jsx?

packages/react-ui-core/src/Form/Email.js
tadjohnston: Yeah, it's still available, but it seems a little weird to have to do `placeholder: ''` just to get it to go away in ag.js

tadjohnston: ag.js forms don't use them.  I think it should be up to the app to decide.

packages/react-ui-core/src/Form/Form.js
tadjohnston: Which part? The hash true was there before, but without adding `empty: true` it wasn't allowing us to override redux properly. This just gives you the option to define how you want to serialize.

packages/react-ui-core/src/Form/RadioGroup.js
tadjohnston: No, not really. It's the same, under the hood jsx converts to this anwyay.

lfender6445:
Do we ever use uncontrolled components for anything? I cant think of an example

packages/react-ui-core/src/Form/Input.js
```diff
@@ -38,11 +40,84 @@ export default class Input extends Component {
...
+
+    /**
+     * Name of value to use for input field, ex: checked or value
+     */
+    property: PropTypes.string,
+
+    /**
+     * Used to determine if InputMask required
+     */
+    mask: PropTypes.string,
```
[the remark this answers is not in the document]
tadjohnston: Done!

packages/react-ui-core/src/LeadForm/LeadForm.js
```diff
@@ -114,6 +121,7 @@ export default class LeadForm extends Component {
         {...props}
       >
         {this.fields}
+        {children}
```
orionstein: I guess we're just doing this so we can throw stuff into the form?

packages/react-ui-core/src/Form/RadioGroup.js
```diff
@@ -77,13 +116,15 @@ export default class RadioGroup extends Component {
     }
   }
 
-  handleClick = event => {
+  @autobind
+  handleClick(event) {
```
grenti: Nice! Makes it consistent 👍

packages/react-ui-core/src/Form/__tests__/RadioGroup-test.js
```diff
@@ -7,101 +7,46 @@ import theme from './mocks/theme'
 const RadioGroup = ThemedRadioGroup.WrappedComponent
 
 describe('Form/RadioGroup', () => {
+  const setup = (props, mountType = shallow) => (
+    mountType(
```
grenti: 👍

packages/react-ui-core/src/LeadForm/LeadForm.js
tadjohnston: Yeah... unfortunately there are things that are difficult to derive like the lead message to have it passed directly into the `fields` prop. It also allows us to add things like a wrapper around button because mobile has a wrapper div (in white) that is positioned absolute on the bottom of the viewport.

tadjohnston: We could also just create a custom component that does this instead, but this just gives you different options.

PR #378, opened 2018-08-22
fix(MoreOptions): replace generic button

LinnJS:

Replace generic buttons with ApplyButton and CancelButton and their concerning test

Reason for change:
Will reduce default styling for this PR
https://github.com/rentpath/ag.js/pull/4947/files

### DIFF
--- a/packages/react-ui-ag/src/Modals/MoreOptionsModal.js
+++ b/packages/react-ui-ag/src/Modals/MoreOptionsModal.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types'
 import themed from 'react-themed'
 import autobind from 'autobind-decorator'
 import classnames from 'classnames'
-import { Modal, Button, Text } from '@rentpath/react-ui-core'
+import { Modal, Button, Text, ApplyButton, CancelButton } from '@rentpath/react-ui-core'
 
 @themed(/^MoreOptionsModal/, { pure: true })
 
@@ -119,18 +119,18 @@ export default class MoreOptionsModal extends PureComponent {
           Reset All
         </Button>
         <div className={theme.MoreOptionsModal_FooterButtons}>
-          <Button
+          <CancelButton
             className={theme.MoreOptionsModal_Cancel}
             onClick={this.handleCancel}
           >
             Cancel
-          </Button>
-          <Button
+          </CancelButton>
+          <ApplyButton
             className={theme.MoreOptionsModal_Submit}
             onClick={this.handleSubmit}
           >
             {submitButton}
-          </Button>
+          </ApplyButton>
         </div>
       </div>
     )
--- a/packages/react-ui-core/src/Button/ApplyButton.js
+++ b/packages/react-ui-core/src/Button/ApplyButton.js
@@ -12,15 +12,15 @@ import Button from './Button'
 export default class ApplyButton extends PureComponent {
   static propTypes = {
     className: PropTypes.string,
+    children: PropTypes.node,
     onClick: PropTypes.func,
     theme: PropTypes.object,
-    name: PropTypes.node,
     value: PropTypes.any,
   }
 
   static defaultProps = {
+    children: 'Apply',
     theme: {},
-    name: 'Apply',
   }
 
   @autobind
@@ -33,9 +33,9 @@ export default class ApplyButton extends PureComponent {
   render() {
     const {
       theme,
-      name,
       className,
       onClick,
+      children,
       value,
       ...props
     } = this.props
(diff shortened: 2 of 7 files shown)


packages/react-ui-ag/src/Modals/MoreOptionsModal.js
```diff
@@ -119,18 +119,16 @@ export default class MoreOptionsModal extends PureComponent {
...
         <div className={theme.MoreOptionsModal_FooterButtons}>
-          <Button
+          <CancelButton
             className={theme.MoreOptionsModal_Cancel}
             onClick={this.handleCancel}
-          >
-            Cancel
-          </Button>
-          <Button
+            name={'Cancel'}
```
tadjohnston: I really think we need to remove the name prop from the CancelButton and ApplyButton. It should just pass in children.

LinnJS: Okie Dokie, I wasn't sure which one to do

packages/react-ui-core/src/Button/CancelButton.js
```diff
@@ -50,9 +49,7 @@ export default class CancelButton extends PureComponent {
         onClick={this.handleClick}
         data-tid="cancel-button"
         {...props}
-      >
-        {name}
```
tadjohnston: You forgot to remove `name` prop.

tadjohnston:
You'll need to check all of react-ui-ag and -react-ui-core if `name` prop is being removed 😄

packages/react-ui-core/src/Button/ApplyButton.js
```diff
@@ -52,7 +52,7 @@ export default class ApplyButton extends PureComponent {
         data-tid="apply-button"
         {...props}
       >
-        {name}
```
ostlerp: Are there any concerns about breaking anything in rent or ag if they are passing `name` instead of `text`?

LinnJS: Yeah I am dumb I will revert to name, I just didnt like the prop name tbh

tadjohnston: I'd prefer it be children instead, but we do need to check where it's being used.  If we tag the BREAKING CHANGES section through `git cz`, then we should be fine 👍

LinnJS: Children it is 🎉

ostlerp: children makes way more sense for this

packages/react-ui-core/src/Button/ApplyButton.js
```diff
@@ -12,15 +12,15 @@ import Button from './Button'
 export default class ApplyButton extends PureComponent {
   static propTypes = {
     className: PropTypes.string,
+    children: PropTypes.string,
```
ostlerp: Since this is a react thing, shouldn't it be `node`?

LinnJS: done && this.done

packages/react-ui-core/src/Button/__tests__/CancelButton-tests.js
```diff
@@ -23,7 +23,8 @@ describe('ag/Buttons/CancelButton', () => {
   })
 
   it('allows an override of the name', () => {
-    const wrapper = shallow(<CancelButton name="Do not cancel!!!" onClick={() => null} />)
+    const children = 'Do not cancel!!!'
+    const wrapper = shallow(<CancelButton onClick={() => null}>{children}</CancelButton>)
     expect(wrapper.find(Button).children().text()).toEqual('Do not cancel!!!')
```
tadjohnston: Should probably say `toEqual(children)`, but this is fine for now

packages/react-ui-core/src/Button/__tests__/ApplyButton-tests.js
```diff
@@ -23,7 +23,8 @@ describe('ag/Buttons/ApplyButton', () => {
   })
 
   it('allows an override of the name', () => {
-    const wrapper = shallow(<ApplyButton name="Do not apply!!!" onClick={() => null} />)
+    const children = 'Do not apply!!!'
+    const wrapper = shallow(<ApplyButton onClick={() => null}>{children}</ApplyButton>)
     expect(wrapper.find(Button).children().text()).toEqual('Do not apply!!!')
```
tadjohnston: Should probably say toEqual(children), but this is fine for now

PR #383, opened 2018-08-28
version bump: react-ui calendar

cwklausing:
Updated the [react-basic-datepicker](https://github.com/rentpath/react-basic-datepicker) in response to below card. Requires a version bump. 

[Card](https://rentpath.leankit.com/card/697016601)

### DIFF
--- a/packages/react-ui-core/package.json
+++ b/packages/react-ui-core/package.json
@@ -53,7 +53,7 @@
     "classnames": "^2.2.6",
     "form-serialize": "0.7.2",
     "prop-types": "^15.6.1",
-    "react-basic-datepicker": "rentpath/react-basic-datepicker",
+    "react-basic-datepicker": "rentpath/react-basic-datepicker#3ca6bd52e6e55419",
     "react-image-gallery": "rentpath/react-image-gallery",
     "react-input-mask": "^2.0.3",
     "react-input-range": "1.3.0",


packages/react-ui-core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@rentpath/react-ui-core",
-  "version": "15.2.1",
+  "version": "15.2.2",
```
tadjohnston: This doesn't bump the datepicker. You'll need to add the datepicker commitish on this line: https://github.com/rentpath/react-ui/pull/383/files#diff-381a8d04c2e36ba268c0723c3a9ce2c5L56

tadjohnston:
Please use `git cz` to commit.

packages/react-ui-core/package.json
cwklausing: @tadjohnston That's the part I was confused about--it's just a reference to the github repo, so how to I show that it's a new version?

cwklausing: 👍 updated

tadjohnston: You still don't need this line 😄 . Publishing does this automatically for you.

cwklausing: ...oh. Gotcha. ha

PR #384, opened 2018-08-28
fix(datepicker): fixes broken ssr

tadjohnston:

* add check to only render on mount so it doesn't fail on SSR
* removed placeholder from
MoveInDate


lfender6445:
👍

PR #389, opened 2018-08-28
fix(datepicker): bump datepicker package

cwklausing:

• bumping datepicker package


tadjohnston:
Let's try this again...

PR #392, opened 2018-08-30
feat(datepicker): clone into repo

tadjohnston:

* moved repo into react-ui-core since it's only 2 files


lfender6445:
I'm good w this but could I ask for storybook entry?

PR #393, opened 2018-09-04
feat(Markers): grey pins for isBasic listings

paulpostel:
[Card](https://rentpath.leankit.com/card/716386936)

All isBasic listings should now receive the grey pin, rather than just the !isActive listings.  isBasic listings are composed of tplsource === 'INACTIVE' + tplsource === 'LNP' (aka "Large Never Paid"). The LNP listings are a new set being added by the inventory team.


tadjohnston:
Please use `git cz` to format your commit properly.

PR #395, opened 2018-09-06
feat(date-fns): upgrading package

tadjohnston:

Fixes a ton of issues with the react-basic-datepicker
* moves it to 2.x (exports are named differently
* clean up of value in datepicker. there is now only one source of truth for both the datepicker selected date and the input value
* add a lot more test coverage and fix other tests
* added react-themed to calendar
* made the input an `<Input />` from react-ui-core
* add ability to format the input value based on the format prop from typing in a valid date in the input field that doesn't match
* added `classnames` to the Calendar so it no longer puts a space on the end of the classname when only one class
* removed `DateInput`. now using the `DatePicker` directly instead.

[Card](https://rentpath.leankit.com/card/705992503)


tadjohnston:
Coverage for the 2 files added:
```
 src/DatePicker           |    96.92 |    95.45 |    93.75 |    96.77 |                   |
  Calendar.js             |      100 |      100 |      100 |      100 |                   |
  DatePicker.js           |    94.29 |    92.86 |    88.89 |    93.94 |             53,58 |
  index.js                |      100 |      100 |      100 |      100 |                   |
```

packages/react-ui-core/package.json
```diff
@@ -40,7 +40,7 @@
     "babel-plugin-transform-runtime": "6.23.0",
     "babel-preset-env": "^1.7.0",
     "babel-preset-react": "6.24.1",
-    "date-fns": "^1.29.0",
+    "date-fns": "2.0.0-alpha.16",
```
mpelzsherman: This is a newer version than the one we're using in AG.js. Do we really need 2 different versions?

packages/react-ui-core/src/DatePicker/DatePicker.js
```diff
@@ -79,72 +58,92 @@ export default class DatePicker extends Component {
     document.removeEventListener('click', this.onOutsideClick, false)
   }
 
+  @autobind
   onOutsideClick(event) {
-    if (!this.mainNode.contains(event.target) && this.state.showCalendar) {
+    if (!this.picker.current.contains(event.target) && this.state.showCalendar) {
```
mpelzsherman: Will `picker.current` always exist?

stories/core/DatePicker.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react'
+import { action } from '@storybook/addon-actions'
+import { DatePicker } from 'react-ui-core/src'
+
+export const DefaultDatePicker = (
+  <DatePicker
+    value="10/10/2018"
+    onClick={action('picked a date')}
```
grenti: I noticed that onClick sends the entire `HtmlInputElement` class to the action logger. Perhaps `onChange` is better because it'll display the `selectedDate` that was passed in `DatePicker.dateChange`?

packages/react-ui-core/src/DatePicker/__tests__/Calendar-test.js
```diff
@@ -1,110 +1,224 @@
...
 
-  it('date can be formatted by prop', () => {
-    const calendar = shallow(<Calendar dateFormat="YYYY-MM-DD" />)
-    expect(format(calendar.state().date, 'YYYY-MM-DD')).toEqual(format(new Date(), 'YYYY-MM-DD'))
+  describe('dateFormat', () => {
+    it('uses default `dateFormat` when none passed', () => {
+      const wrapper = setup()
+      const date = wrapper.state().date
+      const dateFormat = wrapper.instance().props.dateFormat
+      expect(format(date, dateFormat)).toEqual(format(new Date(), dateFormat))
```
mpelzsherman: I'm not sure this test does what it says it does. Shouldn't it be

```javascript
expect(format(date)).toEqual(format(date, DATE_FORMAT))
```

??

stories/core/DatePicker.js
```diff
@@ -0,0 +1,28 @@
...
+    placeholderText="Date"
+  />
+)
+
+export const DatePickerReadOnly = (
+  <DatePicker
+    value="11/01/2018"
+    onClick={action('picked a date')}
+    placeholderText="Date"
```
grenti: nitpick: I think you meant `placeholder` - react-dom is complaining about this after getting spread on the input field.

packages/react-ui-core/src/DatePicker/__tests__/Calendar-test.js
```diff
@@ -1,110 +1,224 @@
...
+    })
+
+    it('uses custom `dateFormat` when no passed', () => {
```
mpelzsherman: "When no passed"?

packages/react-ui-core/src/DatePicker/__tests__/DatePicker-test.js
```diff
@@ -1,86 +1,278 @@
...
+
+          expect(wrapper.state('value')).toEqual(date)
+          expect(wrapper.state('isValidDate')).toBe(true)
+        })
+      })
+    })
 
-    expect(input.props().readOnly).toBeTruthy()
+    it('does not format the `value` and sets `isValidDate` to false when value is not valid', () => {
+      const value = '111-111-11'
```
mpelzsherman: What do you think about using more of an "edge casey" value, like '1/32/2018'?

packages/react-ui-core/src/DatePicker/__tests__/Calendar-test.js
orionstein: Also, why are we testing this? Are we testing something on our end, or are we retesting date-fn/format?

packages/react-ui-core/package.json
tadjohnston: It will get upgraded once this is merged and published.  This is a dev dep here. It's listing `^2.0.0` as a peer dep.

packages/react-ui-core/src/DatePicker/DatePicker.js
tadjohnston: Yes... it's a reference.  That's the way you reference a react component that's been mounted.

packages/react-ui-core/src/DatePicker/__tests__/DatePicker-test.js
tadjohnston: That's fine, I can do that 👍

stories/core/DatePicker.js
tadjohnston: You're right! Not sure how that got in there 🤦‍♂️

packages/react-ui-core/src/DatePicker/__tests__/Calendar-test.js
tadjohnston: @mpelzsherman, this is testing against the default prop if none passed.  The format for the default is _not_ the DATE_FORMAT at the top of the screen.  

@orionstein we're just making sure no one removes the default `format` prop or modifies it. It might be a little overkill... 🤷‍♂️

stories/core/DatePicker.js
tadjohnston: Yeah, this should _really_ be `onChange` in the story...

packages/react-ui-core/package.json
mpelzsherman: You mean AG will get upgraded, I assume.

packages/react-ui-core/src/DatePicker/__tests__/Calendar-test.js
mpelzsherman: It's confusing to me because you're saying "if none passed", but it looks like you ARE passing in a dateFormat. Or maybe you mean "if no **date** is passed"? It's ambiguous. In either case, wouldn't it be clearer to pass in `null` or `undefined` explicitly, rather than forcing the reader to confirm that the wrapper data is undefined? The same applies to the 'date state' tests above BTW. It's just a suggestion, but I think this would make the tests more readable.

PR #397, opened 2018-09-11
chore(packages): upgrades to fix broken unit tests

tadjohnston:
@rentpath/react-ui-tracking, @rentpath/react-ui-utils

Fix broken unit tests with jest


tadjohnston:
There is an issue where coverage isn't reporting correctly.  I can't seem to find anything about this in jest issues, but I really don't want to spend more time on trying to fix it right now since it seems like something they did anyway.

PR #403, opened 2018-09-18
fix(OverlayView): fix race condition of container initialization

ostlerp:

[Card](https://rentpath.leankit.com/card/722560743)
Fix race condition where OverlayView would not render after initialization

### DIFF
--- a/packages/react-ui-core/src/Gmap/OverlayView.js
+++ b/packages/react-ui-core/src/Gmap/OverlayView.js
@@ -21,6 +21,13 @@ export default class OverlayView extends PureComponent {
     theme: {},
   }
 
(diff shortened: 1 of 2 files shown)


ostlerp:
> This looks good, but wondering if there's an easy way to test for race condition failure/success?

Good question! I added a test to verify that the state is getting updated, but since it doesn't really render like a traditional component I'm not sure how else to test this. I'm open to suggestions if there are ideas to cover this better!

ostlerp:
Based on some feedback from @tadjohnston I've added a check for the `state.initialized` value in the render function so its a little more clear why there's a state value for it

PR #406, opened 2018-09-27
Use setState callback to ensure state despite asynchrony

jamessral:
Depending on `this.state` before and after the call can cause a subtle bug (in this case calling the onVisibilityChange callback twice when a parent component calls `setState`). Using the callback syntax to set state and the after set callback ensures that everything happens synchronously and the values in `this.state` are what we expect.

#### **Edit:** This is the real issue below

It is possible with event queueing that the caught event is that of the immediate child.
Include a
new and specific data attribute to reliably check to see if the Dropdown or the anchor is the target
of the click.
Also clean up setState logic in toggleVisibility to not rely on state before and
after setState call.

### DIFF
--- a/packages/react-ui-core/src/Dropdown/Dropdown.js
+++ b/packages/react-ui-core/src/Dropdown/Dropdown.js
@@ -54,21 +54,25 @@ export default class Dropdown extends Component {
     const { onVisibilityChange, toggleOnSelect } = this.props
     const visible = toggleOnSelect ? !this.state.visible : true
 
-    this.setState({ visible })
     if (this.state.visible !== visible) {
       onVisibilityChange(visible)
+      this.setState({ visible })
     }
   }
 
   @autobind
-  changeVisibility(visible) {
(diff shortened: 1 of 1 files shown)


tadjohnston:
Other than the one comment, 👍

jamessral:
@gabepages @WinstonMain Sorry, I didn't see the reviews from y'all. I'm still checking this out by linking it to rent-js and my premature rebase got rid of your comments

packages/react-ui-core/src/Dropdown/Dropdown.js
```diff
@@ -54,21 +54,22 @@ export default class Dropdown extends Component {
...
     const visible = toggleOnSelect ? !this.state.visible : true
 
-    this.setState({ visible })
     if (this.state.visible !== visible) {
       onVisibilityChange(visible)
+      this.setState({ visible })
     }
   }
 
-  @autobind
```
jamessral: This was unused

PR #411, opened 2018-10-04
fix(Dropdown): add conditional CSS className for Dropdown

jamessral:

add conditional CSS className for Dropdown, expand or collapse

### DIFF
--- a/packages/react-ui-core/src/Dropdown/Dropdown.js
+++ b/packages/react-ui-core/src/Dropdown/Dropdown.js
@@ -102,7 +102,7 @@ export default class Dropdown extends Component {
   render() {
     const {
       theme,
-      visible,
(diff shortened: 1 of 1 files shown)


packages/react-ui-core/src/Dropdown/Dropdown.js
```diff
@@ -111,18 +111,21 @@ export default class Dropdown extends Component {
       ...props
     } = this.props
 
+    const isVisible = this.state.visible
```
ninomichaels: could also just destructure this and use `visible ? 'expand' ... ` ? but nbd

jamessral: There is already one called `visible` from props.

PR #412, opened 2018-10-05
feat(Markers): zIndex aptguide map pins above other pins

ostlerp:

[Card](https://rentpath.leankit.com/card/740371798)
Based on the source of a property, zindex the map pins appropriately

### DIFF
--- a/packages/react-ui-ag/src/Gmap/Markers.js
+++ b/packages/react-ui-ag/src/Gmap/Markers.js
@@ -39,6 +39,8 @@ export default class Markers extends PureComponent {
     return feature => {
       const isBasic = get(feature, 'properties.isBasic', false)
       const scale = get(feature, 'properties.iconScale', 1)
+      const source = get(feature, 'properties.source', '')
+      const zIndex = typeof source === 'string' && source.toLowerCase() === 'aptguide' ? 2 : 1
       const icon = !isBasic ? markerIcon : markerInactiveIcon
 
       return {
@@ -52,7 +54,7 @@ export default class Markers extends PureComponent {
           if (onMouseOut) onMouseOut(marker)
         },
         ...this.props.marker(feature),
-        zIndex: !isBasic ? 2 : 1,
+        zIndex,
       }
     }
   }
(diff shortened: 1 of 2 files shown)


packages/react-ui-ag/src/Gmap/Markers.js
```diff
@@ -39,6 +39,8 @@ export default class Markers extends PureComponent {
     return feature => {
       const isBasic = get(feature, 'properties.isBasic', false)
       const scale = get(feature, 'properties.iconScale', 1)
+      const source = get(feature, 'properties.source', '')
+      const zIndex = typeof source === 'string' && source.toLowerCase() === 'aptguide' ? 2 : 1
```
mpelzsherman: Is the `typeof` check really needed here? I thought `===` took care of that.

ostlerp: `toLowerCase` is a string specific function, which is why I wanted to make sure it was a string first

packages/react-ui-ag/src/Gmap/Markers.js
```diff
@@ -52,7 +54,7 @@ export default class Markers extends PureComponent {
           if (onMouseOut) onMouseOut(marker)
         },
         ...this.props.marker(feature),
-        zIndex: !isBasic ? 2 : 1,
```
mpelzsherman: Is the `isBasic` logic no longer needed?

mpelzsherman: NM, I see that's the whole point of the PR!

ostlerp: haha `isBasic` is still used to determine whether its a red or a grey icon

packages/react-ui-ag/src/Gmap/Markers.js
mpelzsherman: Would this be even better, perhaps?

```javascript
  typeof source.toLowerCase === 'function'
```

ostlerp: If source is `undefined` that will cause a type error. "Cannot read property toLowerCase of undefined"

PR #415, opened 2018-10-23
fix(photo): ssr missing img tag

tadjohnston:

Issue: We render the \`\` tag with SSR before. Bbut what happens is the DOM img is loaded in
concurrently while the client bundle is loaded in. So trying to use \`onError\` which react intercepts
ends up in a race condition where the img throws an error in the dom first potentially.
Fix:
Offload setting of src to the handler. Check for differences in src !== fallbackUrl and make
adjustments accordingly. State has to be fired if these 2 are the same so the mounted component will
update and force a re-render.

### DIFF
--- a/packages/react-ui-core/src/Photo/Photo.js
+++ b/packages/react-ui-core/src/Photo/Photo.js
@@ -32,15 +32,18 @@ export default class Photo extends PureComponent {
   constructor(props) {
     super(props)
 
+    this.img = React.createRef()
+
     this.state = {
       url: this.props.url,
-      error: false,
     }
   }
 
   componentDidMount() {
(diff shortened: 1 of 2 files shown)


packages/react-ui-core/src/Photo/Photo.js
```diff
@@ -98,9 +101,21 @@ export default class Photo extends PureComponent {
     )
   }
 
+  get isFallback() {
```
ninomichaels: I believe this would be the same:
```suggestion
  get isFallback() {
    return this.img.current && (this.img.current.src === this.props.fallbackUrl)
  }
```

tadjohnston: 👍

PR #420, opened 2018-11-01
fix(es): wrong env preset

tadjohnston:
@rentpath/react-ui-tracking, @rentpath/react-ui-utils

IE11 didn't like the preset for some reason still. This one works and better than what was previously here.


tadjohnston:
Merging as the current was broken in IE11 anyway.

PR #422, opened 2018-11-02
feat(carousel): add thumbs option

cwklausing:
[Card](https://rentpath.atlassian.net/browse/AL-35)

Creates prop that gives ability to add thumbnail previews to `react-image-gallery`.

This PR is a dependency for [this](https://github.com/rentpath/ag.js/pull/5648) PR

### DIFF
--- a/packages/react-ui-core/src/Carousel/Carousel.js
+++ b/packages/react-ui-core/src/Carousel/Carousel.js
@@ -29,6 +29,7 @@ export default class Carousel extends Component {
     selectedIndex: PropTypes.number,
     onSlide: PropTypes.func,
     showNav: PropTypes.bool,
+    showThumbs: PropTypes.bool,
     navigation: PropTypes.shape({
       previous: PropTypes.oneOfType([
         PropTypes.node,
@@ -60,6 +61,7 @@ export default class Carousel extends Component {
 
   static defaultProps = {
     showNav: false,
+    showThumbs: false,
     lazyLoad: false,
     theme: {},
   }
@@ -197,6 +199,7 @@ export default class Carousel extends Component {
       items,
       navigation,
       pagination,
+      showThumbs,
       ...rest
     } = this.props
 
@@ -212,11 +215,12 @@ export default class Carousel extends Component {
           items={this.items}
           renderItem={this.renderItem}
           onSlide={this.onSlide}
-          showThumbnails={false}
+          showThumbnails={showThumbs}
           showPlayButton={false}
           showFullscreenButton={false}
           startIndex={selectedIndex}
           infinite={false}
+          lastThumbnailOffset={showThumbs ? 3 : null}
           swipeThreshold={10}
           preventDefaultTouchmoveEvent
           {...rest}
--- a/packages/react-ui-core/src/Carousel/PhotoCarousel.js
+++ b/packages/react-ui-core/src/Carousel/PhotoCarousel.js
@@ -21,6 +21,7 @@ export default class PhotoCarousel extends PureComponent {
       }),
     ),
     server: PropTypes.string.isRequired,
+    showThumbs: PropTypes.bool,
     isBackgroundImage: PropTypes.bool,
     lazyLoad: PropTypes.oneOfType([
       PropTypes.bool,
@@ -74,6 +75,7 @@ export default class PhotoCarousel extends PureComponent {
       theme,
       items,
       lazyLoad,
+      showThumbs,
       ...rest
     } = this.props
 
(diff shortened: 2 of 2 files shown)


mpelzsherman:
Looks good! Do we need to add any tests?

cwklausing:
I'm not sure... since we're just passing props (which default to false), I'm inclined to say no. But would also be easy to add a quick prop check if we were interested.

tadjohnston:
Why is this needed? From the looks of it, `...rest` should handle this without needing to add the right right?

tadjohnston:
Why is this needed? From the looks of it, `...rest` should handle this without needing to add the right right? You would just need to pass `showThumbnails` instead.

packages/react-ui-core/src/Carousel/Carousel.js
```diff
@@ -212,11 +215,12 @@ export default class Carousel extends Component {
           items={this.items}
           renderItem={this.renderItem}
           onSlide={this.onSlide}
-          showThumbnails={false}
+          showThumbnails={showThumbs}
```
barelyfitz: Why not name the prop `showThumbs`? Seems like having two different names just confuses things.

cwklausing: fair enough--I think that makes sense.

packages/react-ui-core/src/Carousel/Carousel.js
```diff
@@ -212,11 +215,12 @@ export default class Carousel extends Component {
...
           showPlayButton={false}
           showFullscreenButton={false}
           startIndex={selectedIndex}
           infinite={false}
+          lastThumbnailOffset={showThumbs ? 3 : null}
```
tadjohnston: What if we wanted 4? How does that work here?

mpelzsherman:
@tadjohnston I get the reluctance to add unnecessary props, but I think a `showThumbnails` prop makes sense on `Carousel`. Exposing props in the component where they are most likely to be used is a lot more developer-friendly.

cwklausing:
In light of Tad's comments about the `...rest`, I think we can achieve anything we need here without altering React-UI

tadjohnston:
@mpelzsherman I disagree. It's not ideal to keep adding one-off props to appease a use case in one repo.  The fact that the component here can already handle the prop in question means it's flexible enough to handle what's thrown at it which was the original intention of writing components in this library in this manner.  If we were to add a prop in every situation, this would become a mess and also require a publish / release every single time.  

These are all the props react-image-gallery currently takes in:
```
    flickThreshold: PropTypes.number,
    items: PropTypes.array.isRequired,
    showNav: PropTypes.bool,
    autoPlay: PropTypes.bool,
    lazyLoad: PropTypes.bool,
    infinite: PropTypes.bool,
    showIndex: PropTypes.bool,
    showBullets: PropTypes.bool,
    showThumbnails: PropTypes.bool,
    showPlayButton: PropTypes.bool,
    showFullscreenButton: PropTypes.bool,
    disableThumbnailScroll: PropTypes.bool,
    disableArrowKeys: PropTypes.bool,
    disableSwipe: PropTypes.bool,
    useBrowserFullscreen: PropTypes.bool,
    preventDefaultTouchmoveEvent: PropTypes.bool,
    defaultImage: PropTypes.string,
    indexSeparator: PropTypes.string,
    thumbnailPosition: PropTypes.string,
    startIndex: PropTypes.number,
    slideDuration: PropTypes.number,
    slideInterval: PropTypes.number,
    swipeThreshold: PropTypes.number,
    swipingTransitionDuration: PropTypes.number,
    onSlide: PropTypes.func,
    onScreenChange: PropTypes.func,
    onPause: PropTypes.func,
    onPlay: PropTypes.func,
    onClick: PropTypes.func,
    onImageLoad: PropTypes.func,
    onImageError: PropTypes.func,
    onTouchMove: PropTypes.func,
    onTouchEnd: PropTypes.func,
    onTouchStart: PropTypes.func,
    onMouseOver: PropTypes.func,
    onMouseLeave: PropTypes.func,
    onThumbnailError: PropTypes.func,
    onThumbnailClick: PropTypes.func,
    renderCustomControls: PropTypes.func,
    renderLeftNav: PropTypes.func,
    renderRightNav: PropTypes.func,
    renderPlayPauseButton: PropTypes.func,
    renderFullscreenButton: PropTypes.func,
    renderItem: PropTypes.func,
    stopPropagation: PropTypes.bool,
    additionalClass: PropTypes.string,
    useTranslate3D: PropTypes.bool,
    isRTL: PropTypes.bool
```

So where do you stop in adding these? Is it very developer friendly to spread props like we're doing? Probably not, but I don't see how that's any different than having to track a prop (prop drilling) through multiple levels possibly.

Unless you're specifically adding logic for the prop coming in or you're doing something weird with props where you need to desctructure it, there generally isn't a good use case to add it IMO.

PR #430, opened 2018-11-15
feat(Add new icon for gmap): add an icon to use for the pdp map

bethstevic:

### DIFF
--- a/packages/react-ui-ag/src/Gmap/Markers.js
+++ b/packages/react-ui-ag/src/Gmap/Markers.js
@@ -1,30 +1,27 @@
 import React, { PureComponent } from 'react'
 import PropTypes from 'prop-types'
 import get from 'lodash/get'
+import keys from 'lodash/keys'
 import { Markers as Gmarkers } from '@rentpath/react-ui-core'
-import {
-  redDotIcon,
-  greyDotIcon,
-  blackDotIcon,
-} from './markerIcons'
+import MarkerIcons from './markerIcons'
 
 const NOOP = () => ({})
 
 export default class Markers extends PureComponent {
   static propTypes = {
     marker: PropTypes.func,
-    markerIcon: PropTypes.func,
-    markerInactiveIcon: PropTypes.func,
-    markerIconHover: PropTypes.func,
+    markerIcon: PropTypes.oneOf(keys(MarkerIcons)),
+    markerInactiveIcon: PropTypes.oneOf(keys(MarkerIcons)),
+    markerIconHover: PropTypes.oneOf(keys(MarkerIcons)),
     onMouseOver: PropTypes.func,
     onMouseOut: PropTypes.func,
   }
 
   static defaultProps = {
     marker: NOOP,
-    markerIcon: redDotIcon,
-    markerInactiveIcon: greyDotIcon,
-    markerIconHover: blackDotIcon,
+    markerIcon: 'redDotIcon',
+    markerInactiveIcon: 'greyDotIcon',
+    markerIconHover: 'blackDotIcon',
   }
 
   get marker() {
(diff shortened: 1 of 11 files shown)


packages/react-ui-ag/src/Gmap/markerIcons.js
```diff
@@ -6,26 +6,45 @@ const scaledSize = (x, y, scale) => (
...
+
+const redPdpMapBalloon = (scale = 1) => ({
+  url: "data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' width='28' height='40' viewBox='0 0 28 40'%3E%3Cg fill='none' fill-rule='evenodd'%3E%3Cpath fill='%23FF132F' fill-rule='nonzero' d='M14 0C6.27 0 0 6.27 0 14c0 10.5 14 26 14 26s14-15.5 14-26c0-7.73-6.27-14-14-14z'/%3E%3Cellipse cx='14' cy='12.982' fill='%23FFF' rx='5.25' ry='5.263'/%3E%3C/g%3E%3C/svg%3E",
+  anchor: scaledPoint(15, 38, scale),
+})
+
+const redPdpMapBalloonWithBlackDot = (scale = 1) => ({
+  url: "data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' width='28' height='55' viewBox='0 0 28 55'%3E%3Cdefs%3E%3Cfilter id='a' width='100%25' height='100%25' x='0%25' y='0%25' filterUnits='objectBoundingBox'%3E%3CfeGaussianBlur in='SourceGraphic'/%3E%3C/filter%3E%3C/defs%3E%3Cg fill='none' fill-rule='evenodd'%3E%3Ccircle cx='14' cy='48' r='7' fill='%23000' filter='url(%23a)'/%3E%3Cpath fill='%23FF132F' fill-rule='nonzero' d='M14 0C6.27 0 0 6.27 0 14c0 10.5 14 26 14 26s14-15.5 14-26c0-7.73-6.27-14-14-14z'/%3E%3Cellipse cx='14' cy='12.982' fill='%23FFF' rx='5.25' ry='5.263'/%3E%3C/g%3E%3C/svg%3E%0A",
+  anchor: scaledPoint(16, 48, scale),
+})
```
ostlerp: You may want a `scaledSize` property for these. Its probably optional since it's unlikely you're using it, but it would be necessary for `scale` to be used properly down the road.

packages/react-ui-ag/src/Gmap/Markers.js
```diff
@@ -2,29 +2,25 @@ import React, { PureComponent } from 'react'
...
 
 const NOOP = () => ({})
 
 export default class Markers extends PureComponent {
   static propTypes = {
     marker: PropTypes.func,
-    markerIcon: PropTypes.func,
-    markerInactiveIcon: PropTypes.func,
-    markerIconHover: PropTypes.func,
+    markerIcon: PropTypes.string,
```
ostlerp: Since these are now limited to the icons we have defined, we may want this to be a `oneOf[...]` instead of an open string.

packages/react-ui-ag/src/Gmap/markerIcons.js
bethstevic: Ok, that make sense - I will add in the scaled size property

packages/react-ui-ag/src/Gmap/Markers.js
bethstevic: Good point - I will make that change

packages/react-ui-ag/src/Gmap/Markers.js
```diff
@@ -2,29 +2,34 @@ import React, { PureComponent } from 'react'
...
+
 export default class Markers extends PureComponent {
   static propTypes = {
     marker: PropTypes.func,
-    markerIcon: PropTypes.func,
-    markerInactiveIcon: PropTypes.func,
-    markerIconHover: PropTypes.func,
+    markerIcon: PropTypes.oneOf(markerIcons),
+    markerInactiveIcon: PropTypes.oneOf(markerIcons),
+    markerIconHover: PropTypes.oneOf(markerIcons),
```
ostlerp: I wonder if `_.keys(MarkerIcons)` would work here? That way we wouldn't have to maintain a list of keys

PR #431, opened 2018-11-16
feat(markerIcons): update the color of map pins to new brand red

ostlerp:

[Jira Ticket](https://rentpath.atlassian.net/browse/SANDS-37)

Update ag map pins to match new branding color

#d32526 -> #ff132f


mpelzsherman:
Is it necessary to set the pin color in the SVG? It would be easier to change if we set it in CSS, no?

ostlerp:
> Is it necessary to set the pin color in the SVG? It would be easier to change if we set it in CSS, no?

Seems the best approach to me given the color is already in svg. I'm open to alternatives, but I'm having a hard time imagining a good css solution to this.

bethstevic:
I think usually it is a good idea to manage colors with css, but we are inserting these icons into gmap and we will not have access to them to change the color afterward. In this case, I think we need to define the color in the svg.

PR #433, opened 2018-12-12
Update commitizen to remove flatmap-stream dep

arfordweb:
Commitizen v3.0.5 was released a couple weeks ago to stop the hacked `flatmap-stream` package from getting into its dependency tree.  This PR updates `commitizen` to the latest version so we can take advantage of this security fix.

More info about `event-stream` and `flatmap-stream` https://github.com/dominictarr/event-stream/issues/115 (This overall issue made headlines three weeks ago)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ coverage
 build
 artifacts
 .idea
+.vscode
--- a/yarn.lock
+++ b/yarn.lock
@@ -4955,10 +4955,10 @@ commander@~2.13.0:
   resolved "https://registry.yarnpkg.com/commander/-/commander-2.13.0.tgz#6964bca67685df7c1f1430c584f07d7597885b9c"
   integrity sha512-MVuS359B+YzaWqjCL/c+22gfryv+mCBPHAv3zyVI2GN8EY6IRP8VwtasXn8jyyhvvq84R4ImN1OKRtcbIasjYA==
 
-commitizen@^3.0.4:
-  version "3.0.4"
-  resolved "https://registry.yarnpkg.com/commitizen/-/commitizen-3.0.4.tgz#557904e15cc0641d5b7cdcba2f3e74ef3b693d79"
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -13,3 +13,4 @@ coverage
 build
 artifacts
 .idea
+.vscode
```
arfordweb: VS Code creates this folder.  I added it here so VS Code users don't accidentally commit this folder to the repo.

yarn.lock
```diff
@@ -6517,12 +6517,11 @@ event-emitter@~0.3.5:
     es5-ext "~0.10.14"
 
 event-stream@^3.3.0:
-  version "3.3.6"
-  resolved "https://registry.yarnpkg.com/event-stream/-/event-stream-3.3.6.tgz#cac1230890e07e73ec9cacd038f60a5b66173eef"
-  integrity sha512-dGXNg4F/FgVzlApjzItL+7naHutA3fDqbV/zAZqDDlXTjiMnQmZKu+prImWKszeBM5UQeGvAl3u1wBiKeDh61g==
+  version "3.3.5"
```
arfordweb: `event-stream@3.3.6` is when the dependency on `flatmap-stream` was introduced.  So, projects have downgraded to `event-stream@3.3.5` to do away with the offending dependency.

yarn.lock
```diff
@@ -7037,11 +7036,6 @@ flat-cache@^1.2.1:
     graceful-fs "^4.1.2"
     write "^0.2.1"
 
-flatmap-stream@^0.1.0:
-  version "0.1.1"
-  resolved "https://registry.yarnpkg.com/flatmap-stream/-/flatmap-stream-0.1.1.tgz#d34f39ef3b9aa5a2fc225016bd3adf28ac5ae6ea"
-  integrity sha512-lAq4tLbm3sidmdCN8G3ExaxH7cUCtP5mgDvrYowsx84dcYkJJ4I28N7gkxA6+YlSXzaGLJYIDEi9WGfXzMiXdw==
-
```
arfordweb: Here's the naughty package.

PR #441, opened 2019-01-28
fix(Impacts PhotoCarousel component): improve flexibility by separati…

lfender6445:
https://rentpath.atlassian.net/browse/TIDE-542

improve flexibility by separating lazyload props

Lazyloading the entire gallery is causing problems for us with browser back + forward navigation.
We'd rather just lazy load the photos. This change lets us do either or

BREAKING CHANGE:
Those who wish to continue lazy loading the gallery will need to use prop.lazyLoadGallery=true

ISSUES CLOSED: TIDE-542

### DIFF
--- a/packages/react-ui-core/src/Carousel/__tests__/PhotoCarousel-test.js
+++ b/packages/react-ui-core/src/Carousel/__tests__/PhotoCarousel-test.js
@@ -4,6 +4,7 @@ import LazyLoad from 'react-lazyload'
 import renderer from 'react-test-renderer'
 import theme from './mocks/theme'
 import ThemedPhotoCarousel from '../PhotoCarousel'
+import ThemedCarousel from '../Carousel'
 
 const PhotoCarousel = ThemedPhotoCarousel.WrappedComponent
 const items = [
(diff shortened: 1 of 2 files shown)


tadjohnston:
This line needs to be changed: https://github.com/rentpath/react-ui/pull/441/files#diff-77cfbf6ff3b6489cd8ea45234d1d4823R89. `...rest` can just pass the `lazyLoad` prop through.

tadjohnston:
This is also a problem: https://github.com/rentpath/react-ui/pull/441/files#diff-77cfbf6ff3b6489cd8ea45234d1d4823R81. It will now pass `lazyLoadGallery` through. It needs to be destructured.

packages/react-ui-core/src/Carousel/__tests__/PhotoCarousel-test.js
```diff
@@ -101,9 +102,22 @@ describe('PhotoCarousel', () => {
         theme={theme}
         items={items}
         server="https://rentpath-res.cloudinary.com/"
-        lazyLoad={{ once: true, height: 400, width: 300, offset: 200 }}
+        lazyLoad
       />
     )
-    expect(wrapper.find(<LazyLoad />)).toBeTruthy()
+    expect(wrapper.find(ThemedCarousel).prop('lazyLoad')).toBeTruthy()
```
tadjohnston: This would be better IMO:
```
expect(wrapper.find(ThemedCarousel).prop('lazyLoad')).toBe(true)
```

lfender6445: done

PR #444, opened 2019-03-18
Fix link to react-themed docs.

mpelzsherman:
[Task](http://foo)


mneumark:
@mpelzsherman I saw this when i just merged my PR to change from Ari's react-themed to `@rentpath/react-themed` so I fixed it, so feel free to close this PR. Its taken care of.

PR #445, opened 2019-03-20
fix(Photos.js): add fallback image url to url set in constructor

bradl3yC:


packages/react-ui-core/src/Photo/Photo.js
```diff
@@ -34,7 +34,7 @@ export default class Photo extends PureComponent {
     this.img = React.createRef()
 
     this.state = {
-      url: this.props.url,
+      url: this.props.url || this.props.fallbackUrl,
```
lfender6445: add a test please

PR #450, opened 2019-04-22
feat(Remove Listings code from react-ui-ag)

bethstevic:
we have moved this code into the ag.js code base

[Jira](https://rentpath.atlassian.net/jira/software/projects/TIDE/boards/10?selectedIssue=TIDE-967)


stories/ag/stories.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react'
 import { storiesOf } from '@storybook/react'
 import { withInfo } from '@storybook/addon-info'
-import { Theme } from '@rentpath/react-themed'
+import { Theme } from 'react-themed'
```
tadjohnston: I don't think this will work like this.

bethstevic: I was getting a linting error, but I can put it back as it was if it works the other way.

lfender6445: did you run install? the lib was recently moved to rentpath namespace

lfender6445: your linter was probably failing cause it couldnt find the new dep

packages/react-ui-ag/src/Listings/MobileMapListing.js
```diff
@@ -1,316 +0,0 @@
...
-import themed from '@rentpath/react-themed'
-import classnames from 'classnames'
-import autobind from 'autobind-decorator'
-import get from 'lodash/get'
-import isEqual from 'lodash/isEqual'
-import LazyLoad, { forceCheck } from 'react-lazyload'
-import {
-  Button,
-  ToggleButton,
-  ListingComponents,
```
lfender6445: are we able to also remove ListingComponents and ListingCell from react-ui-core?

lfender6445: hmm ok it looks like ListingComponents is still being used by #967

lfender6445: @m0rd3c4i @quantrells should we move ListingComponents and ListingCell into ag for redesign preperation as well?

mpelzsherman:
JIRA ticket?

mpelzsherman:
Is `react-ui-ag` effectively dead? Will we eventually move all of `react-ui-ag` back into ag?

packages/react-ui-ag/src/Listings/MobileMapListing.js
m0rd3c4i: I'm not sure what our goal was/is with `react-ui`, honestly; it would seem to be even less defined with `react-ui-ag`. I'm on board with moving and building out the new stuff in ag.js.

lfender6445:
yes mps - @bethstevic i am ok if we want to handle the remaining work as a separate card. up to you

packages/react-ui-ag/src/Listings/MobileMapListing.js
quantrells: I agree that LisitingCell will also be a good candidate to move into ag. Its only used by the components in the `react-ui-ag/Lisitngs` folder.

packages/react-ui-core/src/index.js
```diff
@@ -91,11 +91,6 @@ export {
   PhotoCarousel,
 } from './Carousel'
 
-export {
```
tadjohnston: You sure this isn't used by anyone else?

bethstevic: Yes, I checked rent-js and rentals-js and neither were importing these components.

PR #451, opened 2019-04-23
feat(Carousel): Fullbleed Gallery Proptype Support

quantrells:
Updates to proptypes to support slides of fullbleed gallery

### DIFF
--- a/packages/react-ui-core/src/Carousel/PhotoCarousel.js
+++ b/packages/react-ui-core/src/Carousel/PhotoCarousel.js
@@ -7,18 +7,25 @@ import autobind from 'autobind-decorator'
 import { Photo, BackgroundPhoto } from '../Photo'
 import Carousel from './Carousel'
 
+const carouselPhoto = PropTypes.arrayOf(
+  PropTypes.shape({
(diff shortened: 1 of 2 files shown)


packages/react-ui-core/src/Carousel/PhotoCarousel.js
```diff
@@ -12,13 +12,24 @@ export default class PhotoCarousel extends PureComponent {
...
-      }),
-    ),
+    items: PropTypes.oneOfType([
+      PropTypes.arrayOf(
+        PropTypes.arrayOf(
+          PropTypes.shape({
+            caption: PropTypes.string,
+            path: PropTypes.string,
+            url: PropTypes.string,
+          }),
```
mpelzsherman: Can we make this into a type (e.g. `CarouselPhoto`)  instead of copying it?

```js
          PropTypes.shape({
            caption: PropTypes.string,
            path: PropTypes.string,
            url: PropTypes.string,
          }),
```

quantrells: 👍

PR #453, opened 2019-04-26
feat(AutoSuggestField): fixes and new features for Rent implementation of AutoSuggestField

mneumark:

[Story](https://rentpath.atlassian.net/browse/HNL-378)

* add componentRef prop to `Input` to pass ref all the way down to `input`
* add componentRef prop to `Button` to pass ref all the way down to `button`
* switch autobinds to arrow functions in touched files
* changes to AutoSuggestField
  * allow suggestions to be an array of objects
  * added support for feature so selected value populates input
  * fixed highlight styling issues
  * fix highlighted list item issue with onClick handlers
* changes to Highlighter
  * added support for case insensitivity
  * added support for renderer which allows wrapping outputted node
* changes to Menu
  * changed the default index to -1
  * clean up indexes in `Menu`

### DIFF
--- a/packages/react-ui-core/src/Form/Input.js
+++ b/packages/react-ui-core/src/Form/Input.js
@@ -70,6 +70,11 @@ export default class Input extends Component {
      * Used to determine if InputMask required
      */
     mask: PropTypes.string,
+
+    /**
+     * Adds ability to pass ref down to the actual DOM node
+     */
+    componentRef: PropTypes.func,
   }
 
   static defaultProps = {
@@ -128,6 +133,7 @@ export default class Input extends Component {
       className,
       changeEvent,
       property,
+      componentRef,
       value,
       ...props
     } = this.props
@@ -144,6 +150,7 @@ export default class Input extends Component {
           invalid && theme['Input-invalid'],
           variant && theme[`Input-${variant}`],
         )}
+        ref={componentRef}
         {...props}
         {...this.controlledProps}
       />
--- a/packages/react-ui-core/src/Highlighter/Highlighter.js
+++ b/packages/react-ui-core/src/Highlighter/Highlighter.js
@@ -1,51 +1,56 @@
-import React, { Component } from 'react'
+import React, { PureComponent } from 'react'
 import PropTypes from 'prop-types'
 import themed from '@rentpath/react-themed'
 import classnames from 'classnames'
 import { randomId } from '@rentpath/react-ui-utils'
 
 @themed(/^Highlighter/)
-export default class Highlighter extends Component {
+export default class Highlighter extends PureComponent {
   static propTypes = {
     className: PropTypes.string,
     pattern: PropTypes.string.isRequired,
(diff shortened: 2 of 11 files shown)


packages/react-ui-core/src/Form/Input.js
```diff
@@ -70,6 +70,11 @@ export default class Input extends Component {
      * Used to determine if InputMask required
      */
     mask: PropTypes.string,
+
+    /**
+     * Adds ability to pass ref down to the actual DOM node
+     */
+    componentRef: PropTypes.func,
```
tadjohnston: I think you want to use this: https://reactjs.org/docs/forwarding-refs.html#forwarding-refs-in-higher-order-components

packages/react-ui-core/src/Highlighter/Highlighter.js
```diff
@@ -1,51 +1,56 @@
...
   static propTypes = {
     className: PropTypes.string,
     pattern: PropTypes.string.isRequired,
-    children: PropTypes.string,
+    ignoreCase: PropTypes.bool,
+    children: PropTypes.string.isRequired,
     theme: PropTypes.object,
     indexHighlighted: PropTypes.number,
-    text: PropTypes.string,
+    renderer: PropTypes.func,
```
tadjohnston: What is this needed for?

packages/react-ui-core/src/Menu/Menu.js
```diff
@@ -98,11 +93,18 @@ export default class Menu extends PureComponent {
...
 
   get highlightedOption() {
-    return this.options[this.state.highlightIndex || 0]
+    return this.options[this.state.highlightIndex]
   }
 
-  @autobind
-  handleSelection() {
+  get indexedOptions() {
+    return this.options.map((option, index) => ({
```
tadjohnston: Would it not make more sense to use `reduce()` here instead of `.map().filter()`?

packages/react-ui-core/src/Form/Input.js
mneumark: This only sends the ref down one level. We need it to go down about 6 levels or so, so I figured this was cleaner.

packages/react-ui-core/src/Highlighter/Highlighter.js
mneumark: Basically, If you look at how the highlight component is passed down from AutoSuggestField to Menu to List to ListItem, It doesn't exactly leave a lot of room for modifying how its called (in our case we need the ability to prefix an svg to the front of the list item. Given the structure of everything, it was going to involve significant refactors to how stuff is passed down. Or I could just do this. I can walk you through it and see if you have any better ideas.

packages/react-ui-core/src/Menu/Menu.js
mneumark: I can change that. I moved this logic from elsewhere. I didn't write it so I didn't look that closely.

PR #458, opened 2019-05-15
feat(modal): add overlayClassName prop

cwklausing:
[Card](https://rentpath.atlassian.net/browse/INVP-443)

In order to resolve INVP-441 (restoring headless e2e testing) we need to add ability to give the React-UI modal overlay a custom class. Previously, applying universal styling to the overlay in`rent-js` was disrupting some headless e2e tests.

### DIFF
--- a/packages/react-ui-core/src/Modal/Modal.js
+++ b/packages/react-ui-core/src/Modal/Modal.js
@@ -22,6 +22,7 @@ export default class Modal extends PureComponent {
     className: PropTypes.string,
     children: PropTypes.any,
     hasOverlay: PropTypes.bool,
+    overlayClassName: PropTypes.string,
     CloseButton: PropTypes.oneOfType([
       PropTypes.func,
       PropTypes.node,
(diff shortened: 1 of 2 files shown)


tadjohnston:
Can you please explain why this is needed? I'm not understanding from the tickets what the purpose of adding an extra prop does.  The Overlay component already has a className, can't you just use that?

packages/react-ui-core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@rentpath/react-ui-core",
-  "version": "18.2.0",
+  "version": "18.2.1",
```
tadjohnston: There is no reason to bump this here. It's done automatically when we release.

cwklausing: ok, gotcha

tadjohnston: Also, this would be a minor version according to semver
https://semver.org/#spec-item-7

cwklausing:
@tadjohnston Basically, I need the overlay to be slightly transparent for a specific component. I could also do this as a `makeOverlayTransparent` prop or something, but this seemed more dynamic.

PR #461, opened 2019-05-23
chore(Node): upgrade to node 12 TIDE-1135

rgdSolutions:
Updated react-ui to use node 12 and node-sass 4.12, also created .node-version and .nvmrc files.

[CARD](https://rentpath.atlassian.net/browse/TIDE-1135)


lfender6445:
We also need to update travis config https://github.com/rentpath/react-ui/blob/master/.travis.yml#L3

.yarnrc
```diff
@@ -0,0 +1,6 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
lfender6445: more on this here if you're interested https://yarnpkg.com/en/docs/cli/policies

rgdSolutions: Thanks! That was the exact link I used to find out about --> yarn policies set-version

lfender6445: oh i know, was leaving this for people who might wonder how the file originated

PR #464, opened 2019-07-24
feat(Carousel): add ability to override Carousel data-tid

mpelzsherman:

The Carousel component has a hard-coded data-tid, which makes it difficult to write automated tests when there are multiple Carousels on the page.

### DIFF
--- a/packages/react-ui-core/src/Carousel/Carousel.js
+++ b/packages/react-ui-core/src/Carousel/Carousel.js
@@ -24,6 +24,7 @@ const carouselItems = PropTypes.arrayOf(
 export default class Carousel extends Component {
   static propTypes = {
     className: PropTypes.string,
+    'data-tid': PropTypes.string,
     theme: PropTypes.object,
     items: PropTypes.oneOfType([
       PropTypes.arrayOf(
@@ -68,6 +69,7 @@ export default class Carousel extends Component {
     showNav: false,
     lazyLoad: false,
     theme: {},
+    'data-tid': 'carousel',
   }
 
   componentWillReceiveProps(nextProps) {
@@ -208,6 +210,7 @@ export default class Carousel extends Component {
       items,
       navigation,
       pagination,
+      'data-tid': dataTid,
       ...rest
     } = this.props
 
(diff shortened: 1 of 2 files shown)


packages/react-ui-core/src/Carousel/Carousel.js
```diff
@@ -68,6 +69,7 @@ export default class Carousel extends Component {
     showNav: false,
     lazyLoad: false,
     theme: {},
+    'data-tid': 'carousel',
```
lfender6445: i wouldn't default this since the lib is open source, some people might not want the extra data attribute

packages/react-ui-core/src/Carousel/Carousel.js
```diff
@@ -24,6 +24,7 @@ const carouselItems = PropTypes.arrayOf(
 export default class Carousel extends Component {
   static propTypes = {
     className: PropTypes.string,
+    'data-tid': PropTypes.string,
```
lfender6445: this feels like something that could be optionally spread in from the the `...rest` below - any thoughts?

tadjohnston:
I'm not sure why we need this. I'd much rather prefer we keep things like this generic and use a better selector like `[data-tid="some-specific-thing"] [data-tid="carousel"]`.  The reason I would prefer this is because it provides a consistent way of writing tests. I always know that if I'm querying any carousel I don't have to look through code to figure out what the `data-tid` for it is.

packages/react-ui-core/src/Carousel/Carousel.js
tadjohnston: I disagree, unless we don't need this on every carousel, which we may not.  And as open source as this is, I highly doubt anyone else besides RP is using it lol.

packages/react-ui-core/src/Carousel/Carousel.js
tadjohnston: The spread below is passed directly into the Slider and the moment it's destructed to include in the Slider's parent, it has to be declared as a PropType.

mpelzsherman:
Closing based on feedback from @tadjohnston .

PR #465, opened 2019-08-14
feat(react-ui-core image galler): bump version of react-image-gallery…

mpelzsherman:
… to support pre-load divs

Needed for https://rentpath.atlassian.net/browse/INVP-559

### DIFF
--- a/packages/react-ui-core/package.json
+++ b/packages/react-ui-core/package.json
@@ -38,7 +38,7 @@
     "@babel/runtime": "^7.0.0",
     "@rentpath/react-ui-utils": "4.2.7",
     "form-serialize": "0.7.2",
-    "react-image-gallery": "rentpath/react-image-gallery",
+    "react-image-gallery": "rentpath/react-image-gallery#0.8.9",
     "react-input-mask": "2.0.4",
     "react-input-range": "1.3.0",
     "react-lazyload": "2.3.0"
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -12031,14 +12031,16 @@ react-icons@^2.2.7:
...
+  version "0.8.9"
+  resolved "https://registry.yarnpkg.com/react-image-gallery/-/react-image-gallery-0.8.9.tgz#7fc008443d54f0b8b97c208e12a2470d07e26413"
+  integrity sha1-f8AIRD1U8Li5fCCOEqJHDQfiZBM=
   dependencies:
     lodash.debounce "^4.0.8"
     lodash.throttle "^4.1.1"
     prop-types "^15.5.8"
-    react-swipeable "^4.0.0"
+    react-swipeable "^4.2.2"
+    resize-observer-polyfill "^1.5.0"
```
lfender6445: interesting that we even obtained a new dependency. i wonder if this is a side effect of not versioning this thing to begin with

packages/react-ui-core/package.json
```diff
@@ -38,7 +38,7 @@
     "@babel/runtime": "^7.0.0",
     "@rentpath/react-ui-utils": "4.2.7",
     "form-serialize": "0.7.2",
-    "react-image-gallery": "rentpath/react-image-gallery",
+    "react-image-gallery": "0.8.9",
```
lfender6445: hmm should this be a github tag? rentpath/react-image-gallery#0.8.9

yarn.lock
mpelzsherman: We went from `0.8.7` to `0.8.9`, so maybe that did it?

PR #469, opened 2019-11-21
feat(calendar): calendar starts on correct day of week

triedal:
[Card](https://rentpath.atlassian.net/browse/AG-2676)

### DIFF
--- a/packages/react-ui-core/src/DatePicker/Calendar.js
+++ b/packages/react-ui-core/src/DatePicker/Calendar.js
@@ -5,11 +5,16 @@ import autobind from 'autobind-decorator'
 import classnames from 'classnames'
 import isEqual from 'date-fns/isEqual'
 import format from 'date-fns/format'
-import getDaysInMonth from 'date-fns/getDaysInMonth'
 import addMonths from 'date-fns/addMonths'
 import subMonths from 'date-fns/subMonths'
 import isBefore from 'date-fns/isBefore'
 import isAfter from 'date-fns/isAfter'
+import startOfMonth from 'date-fns/startOfMonth'
+import getDay from 'date-fns/getDay'
+import subDays from 'date-fns/subDays'
+import addDays from 'date-fns/addDays'
+import getMonth from 'date-fns/getMonth'
+import getDate from 'date-fns/getDate'
 
 @themed(/^Calendar/, { pure: true })
 export default class Calendar extends PureComponent {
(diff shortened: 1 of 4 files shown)


packages/react-ui-core/src/DatePicker/Calendar.js
```diff
@@ -67,9 +67,9 @@ export default class Calendar extends PureComponent {
     // NOTE: consitently create 5 rows so the calendar pads shorter months
     for (let i = 0; i <= 5; i += 1) {
       days.push(
-        <div key={`date-row-${i}`}>
+        <Fragment key={`date-row-${i}`}>
```
tadjohnston: Why did you change this to a fragment?

triedal: How does this affect styling? It has no classname

tadjohnston: There's no dom element it's attached to. Did you check the output. It probably looks like this based on your change:
```
<div class="Calendar_DaysOfWeek">SuMoTuWeThFriSa</div>
```

tadjohnston: Sorry, I was thinking the wrong thing. But by removing this, it would cause there to be no rows which I assume you could do something with flex to address it, but it might be more difficult than just leaving the div.

packages/react-ui-core/src/DatePicker/Calendar.js
```diff
@@ -43,6 +48,34 @@ export default class Calendar extends PureComponent {
...
+    return daysOfWeek
+  }
+
+  get daysOfMonth() {
+    const days = []
+
+    // NOTE: consitently create 5 rows so the calendar pads shorter months
+    for (let i = 0; i <= 5; i += 1) {
+      days.push(
+        <div key={`date-row-${i}`}>
```
barelyfitz: I don't think there is any need to make the key unique, why not just `key={i}`

barelyfitz: Oh, maybe this array gets concatenated with this other, so you might need that afterall. If so please ignore.

barelyfitz: Since the normal dates of the month are all `button` elements (an inline element) this should probably also be an inline element (`span` instead of a `div`). For example, look at the HTML without styling, you don't want the days to appear one per line, you want them to appear all on the same line like the other dates.

barelyfitz:
Storybook needs to be updated to add styles:
Storybook

packages/react-ui-core/src/DatePicker/Calendar.js
triedal: I believe this should be a block level element as it contains an entire week, not just a day.

tadjohnston: I think you're right on this one @triedal. I would change the days of the week though. I'll try to name some of this stuff better in the ui-ux version and it's somewhat confusing lol.

PR #470, opened 2019-12-13
fix(core): gmap needs to use conventional props

tadjohnston:

We cannot strip the prop types with the current way we use dynamic props. Switching these will fix
the problem


tadjohnston:
It won't matter in production mode as they get stripped out. For some reason, it breaks prod mode right now with webpack in rent.

tadjohnston:
It won't matter in production mode as they get stripped out. For some reason, it breaks prod mode right now with webpack in rent.

PR #476, opened 2020-06-11
Add data-tag_action="ignore" option to checkbox labels

tamars2:
[Jira](https://rentpath.atlassian.net/browse/AG-4378)

- adds option for new data-tag_action
- updates unit test

### DIFF
--- a/packages/react-ui-core/src/Form/Field.js
+++ b/packages/react-ui-core/src/Form/Field.js
@@ -108,6 +108,11 @@ export default class Field extends Component {
       PropTypes.object,
       PropTypes.func,
     ]),
+
+    /**
+     * Ignore tracking on label element.
+     */
+    ignoreAction: PropTypes.bool,
   }
 
   static defaultProps = {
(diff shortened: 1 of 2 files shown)


packages/react-ui-core/src/Form/Field.js
```diff
@@ -199,6 +205,7 @@ export default class Field extends Component {
         key: `label-${inputId}`,
         htmlFor: inputId,
         className: theme.Label,
+        ...(ignoreAction) && { 'data-tag_action': 'ignore' },
```
lfender6445, line 208: does this work for the negative? asking bc i legitimately don't know what happens when you spread `...(false)`

should this be a ternary instead `...(ignoreAction ?{ 'data-tag_action': 'ignore' } : {})`

tamars2: The falsy value should short circuit and wouldn't return the KVP.

See this fiddle for an example: https://jsfiddle.net/vnoktp2u/

I think another way of writing this could be a ternary.  I could go either way, I use lots of ternaries so honestly I was just trying something different here.

tamars2: I supposed the closing brackets should, or could, follow the closing curly bracket.  Seems to work either way in the fiddle above.

PR #480, opened 2020-09-01
feat(RENT-3592): passing through event type on AutoSuggestField actions

RussBrown00:
### [Card](https://rentpath.atlassian.net/browse/RENT-3592)

In order to help dependent code to report the correct actions to google analytics we need to pass through the event type. I started out passing the actual event, but hit issues with React's handling of events as DOM elements were destroyed. This will be fixed in React 17.  I choose my approach to remain backwards compatable.

### DIFF
--- a/packages/react-ui-core/src/AutoSuggestField/AutoSuggestField.js
+++ b/packages/react-ui-core/src/AutoSuggestField/AutoSuggestField.js
@@ -105,7 +105,7 @@ export default class AutoSuggestField extends PureComponent {
         this.handleClear()
         break
       case ENTER:
-        this.handleSubmit(this.state.value)
+        this.handleSubmit(event)
         break
       default:
     }
@@ -166,12 +166,12 @@ export default class AutoSuggestField extends PureComponent {
     }, callback)
   }
 
-  handleSubmit = () => {
-    this.props.onSubmit(this.state.value)
+  handleSubmit = event => {
+    this.props.onSubmit(this.state.value, event.type)
     this.updateValueAndClose(this.state.value)
   }
 
-  handleSuggestionSelection = value => {
+  handleSuggestionSelection = (value, _index, eventType) => {
     const {
       onSubmit,
       onSelection,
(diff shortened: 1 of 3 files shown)


packages/react-ui-core/src/AutoSuggestField/AutoSuggestField.js
```diff
@@ -166,12 +166,12 @@ export default class AutoSuggestField extends PureComponent {
...
 
-  handleSubmit = () => {
-    this.props.onSubmit(this.state.value)
+  handleSubmit = event => {
+    this.props.onSubmit(this.state.value, event.type)
     this.updateValueAndClose(this.state.value)
   }
 
-  handleSuggestionSelection = value => {
+  handleSuggestionSelection = (value, ix, eventType) => {
```
bethstevic, line 174: Do we need `ix`? It doesn't look like it is being used at all

bethstevic (changes requested):
The `enter` event is no longer being fired on click. But, when I do use tab and enter it now fires `keypress`.

packages/react-ui-core/src/AutoSuggestField/AutoSuggestField.js
RussBrown00: We don't need IX for the most part, but to remain backward compatible it's required

RussBrown00:
The only thing this change does is pass along what happened in an extra argument. It won't fix anything until I also update rent to use it. The "old" search is hard wired to submit an 'enter' event

packages/react-ui-core/src/AutoSuggestField/AutoSuggestField.js
mneumark: Can we come up with a better name? I have no idea what ix means

mneumark (approved):
👍 Just one nitpick.

RussBrown00:
> 👍 Just one nitpick.

@mneumark I can change that to something else sure. `ix` and idx` are standard abbreviations for index, but I didn't double check the other code to see another standard first.

mneumark:
I'm not a fan of obfuscation. A lot of those standards are from a bygone 8bit era. Can we just call it "index"?

bethstevic (approved):
We will map `keypress` to the verbiage that the tag team expects client side. I am good with this approach

RussBrown00:
> I'm not a fan of obfuscation. A lot of those standards are from a bygone 8bit era. Can we just call it "index"?

I updated it. even reflected the `_` since it's not being used, even though it's not nessesary.

mneumark (approved):
👍

PR #481, opened 2020-09-02
fix(add-move-in-val): adds move in date field validation

davidthrasher:

This PR adds changes needed to support moveInDate validation for the tour wizard. It coincides with UIUX changes and changes in Rent/AG.

### DIFF
--- a/packages/react-ui-core/src/DatePicker/DatePicker.js
+++ b/packages/react-ui-core/src/DatePicker/DatePicker.js
@@ -19,6 +19,7 @@ export default class DatePicker extends PureComponent {
     onChange: PropTypes.func,
     minDate: PropTypes.object,
     maxDate: PropTypes.object,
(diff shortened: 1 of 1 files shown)


packages/react-ui-core/src/DatePicker/DatePicker.js
```diff
@@ -33,7 +34,7 @@ export default class DatePicker extends PureComponent {
 
     this.picker = React.createRef()
     this.outsideHandler = null
-    this.formatRegex = new RegExp(props.dateFormat.replace('/', '\\/').replace(/[mdy]/ig, '\\d'))
```
davidthrasher, line 36: This wasn't actually creating a valid regex. Decided to hardcode what it was attempting to validate. Which is `mm/dd/yyyy`

gabepages (approved):

PR #482, opened 2020-09-10
fix(ci): address gh_pages deployment failure

kaborso:
[CARD](https://rentpath.atlassian.net/browse/DX-144)

To resolve DX-144, we might only need to update the token.
Convert gh_pages deployment to github actions in another story.


whitneytfoster:
Replaced by PR #487 .

PR #487, opened 2020-09-17
fix: update publish workflow

CraigCottingham:
- rework workflow to match what we’re doing in other workflows
- remove storybook and deploy stages from .travis.yml


.github/workflows/publish.yml
```diff
@@ -1,70 +1,61 @@
...
   publish_packages:
-    runs-on: 'ubuntu-latest'
+    if: github.event.action == 'closed' && github.event.pull_request.merged
+    runs-on: ubuntu-latest
     steps:
-      - name: 'clone repo with full history for lerna'
+      - name: check out latest commit
         uses: actions/checkout@v2
         with:
-          fetch-depth: 0
```
kaborso, line 12: Is `lerna` able to version without this setting?

kaborso: Can we do an example run with `on: pull_request(synchronize)`?

CraigCottingham: Beats me. I know nothing about `lerna`. I _do_ know that packages are published when this workflow is run. I _don't_ know if that one example is sufficiently comprehensive.

CraigCottingham: Regarding a test run: I commented out the `on` criteria and the `if` condition on the `publish_packages` job, and [packages published as expected](https://github.com/rentpath/react-ui/packages).

kaborso: Awesome

kaborso (approved):

tadjohnston (approved):

PR #490, opened 2020-10-14
Remove modal overlay lock on unmount

bethstevic:
- Fixes bug where classname was not removed on modal close

[Jira Bug Ticket](https://rentpath.atlassian.net/browse/AG-4926)

### DIFF
--- a/packages/react-ui-core/src/ModalStack/ModalStack.js
+++ b/packages/react-ui-core/src/ModalStack/ModalStack.js
@@ -56,6 +56,11 @@ export default class ModalStack extends PureComponent {
     this.loadCurrentModal()
   }
 
+  componentWillUnmount() {
+    this.toggleBodyClass(false)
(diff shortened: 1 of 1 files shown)


packages/react-ui-core/src/ModalStack/ModalStack.js
```diff
@@ -109,10 +114,7 @@ export default class ModalStack extends PureComponent {
 
   toggleBodyClass(toggle) {
     const { theme } = this.props
-
-    if (this.modalHost) {
-      document.body.classList.toggle(theme['Overlay-lock'], toggle)
-    }
+    document.body.classList.toggle(theme['Overlay-lock'], toggle)
```
barelyfitz, line 117: I'm curious if this works as expected in IE - according to `toggle()` docs IE does not support second argument: https://developer.mozilla.org/en-US/docs/Web/API/Element/classList

PR #506, opened 2021-07-02
RENT-4856 Update Modal Component for Scroll Locking

tamars2:
[Jira](https://rentpath.atlassian.net/browse/RENT-4856)

- move scroll locking to Modal component
- pass new prop (defaulted to true) to allow for control of scroll locking

### DIFF
--- a/packages/react-ui-core/src/Modal/Modal.js
+++ b/packages/react-ui-core/src/Modal/Modal.js
@@ -22,6 +22,7 @@ export default class Modal extends PureComponent {
     className: PropTypes.string,
     children: PropTypes.any,
     hasOverlay: PropTypes.bool,
+    lockBodyScrolling: PropTypes.bool,
     CloseButton: PropTypes.oneOfType([
       PropTypes.func,
       PropTypes.node,
@@ -34,6 +35,7 @@ export default class Modal extends PureComponent {
     theme: {},
     hasOverlay: true,
     closeOnOverlayClick: true,
+    lockBodyScrolling: true,
   }
 
   constructor(props) {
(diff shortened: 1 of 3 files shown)


packages/react-ui-core/src/Modal/Modal.js
```diff
@@ -93,6 +99,12 @@ export default class Modal extends PureComponent {
     if (this.props.onClose) this.props.onClose()
   }
 
+  toggleBodyClass = toggle => {
```
tamars2, line 102: I moved this responsibility to the Modal component so that Modals themselves can decide whether or not the body should have the `-lock` classname applied to it.

packages/react-ui-core/src/Modal/Modal.js
```diff
@@ -103,9 +115,14 @@ export default class Modal extends PureComponent {
       CloseButton,
       hasOverlay,
       closeOnOverlayClick,
+      lockBodyScrolling,
       ...props
     } = this.props
 
+    if (lockBodyScrolling) {
```
tamars2, line 122: Defaulted to true, pass it as false to avoid applying `Overlay-lock` to `body`.

packages/react-ui-core/src/Modal/Modal.js
```diff
@@ -103,9 +115,14 @@ export default class Modal extends PureComponent {
...
+      this.toggleBodyClass(true)
```
kmorales13, line 123: Is this something that should happen inside the render method or can we move it to a lifecycle method?

tamars2: I could move it to `componentDidMount`, I think.

packages/react-ui-core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@rentpath/react-ui-core",
-  "version": "19.4.3",
+  "version": "19.5.0",
```
joseph-walker, line 3: Does this repo not auto-bump versions on publish?

packages/react-ui-core/src/Modal/Modal.js
joseph-walker: Good idea. Otherwise the DOM will try to update the document root classlist every single time React renders.

kmorales13: Yup, that would make more sense, since it seems it should only happen once

packages/react-ui-core/package.json
tamars2: 🤔 

I need to look further into this.

joseph-walker (dismissed):

joseph-walker (approved):

gabepages (approved):

kmorales13 (approved):

